Issue #148, opened 2014-05-28
Tray icon is displayed in the upper left corner in Ubuntu 14.04


zhoulike:
![tray-icon](https://dl.dropboxusercontent.com/u/22403480/zeal.png)

This issue is similar with [issue 104](https://github.com/jkozera/zeal/issues/104) except that tray icon is not a black box. I'm using Ubuntu 14.04. Zeal is the latest version.

Any solutions?

gausie:
This problem has returned for me.

trollixx:
The problem has been fixed in 0.1 branch (see #321), and will be released as part of 0.1.2 release shortly.

PR #151, opened 2014-06-03
Fix the tray icon problem in Unity

zhoulike:
Tray icon is displayed wrong in Ubuntu 14.04, as shown in [issue 148](https://github.com/jkozera/zeal/issues/148). Then, I use application indicator instead of tray icon in Ubuntu 14.04 following the instructions. 

Note: left-click on application indicator won't show the application. This is the feature of application indicator. We can use hotkey to show the application. 

New requirements: libgtk2.0-dev libappindicator-dev


trollixx:
Referencing #321 & #148 just to link all issues together.

PR #255, opened 2015-01-12
set User-Agent for HTTP downloads

### LINKED ISSUE #250: installing some docsets fails
Some docsets, e.g. Python 2, cannot be installed. They looks as follows after download.

unbenannt

After restart, they are completely gone.

schlamar:
Trying to fix #250, not sure if this works, though :)


trollixx:
Please squash your commits into one. Does it make Zeal work with your proxy setup?

schlamar:
Yes, this resolves download issues for me. Updated the PR.

trollixx:
Great, thanks!

PR #260, opened 2015-01-24
MSVC Fixes

lowjoel:
Because on MSVC Qt uses wchar_t, QStringLiteral can only be applied on one constant string at a time.

### DIFF
--- a/src/ui/settingsdialog.cpp
+++ b/src/ui/settingsdialog.cpp
@@ -396,7 +396,8 @@ void SettingsDialog::processDocsetList(const QJsonArray &list)
 {
     for (const QJsonValue &v : list) {
         QJsonObject docsetJson = v.toObject();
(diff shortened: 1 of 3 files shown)


src/ui/settingsdialog.cpp
```diff
@@ -396,7 +396,8 @@ void SettingsDialog::processDocsetList(const QJsonArray &list)
 {
     for (const QJsonValue &v : list) {
         QJsonObject docsetJson = v.toObject();
-        docsetJson[QStringLiteral("source")] = QStringLiteral("kapeli");
```
trollixx: Out of curiosity, why this one does not work with MSVC? There were two separate `QStringLiteral`s.

lowjoel: QStringLiteral on Windows is a macro expansion to a block.

lowjoel: sorry, it expands to a lambda that is applied.

PR #270, opened 2015-02-10
Add PREFIX and PREFIX_USR_SHARE

phracek:
This pull request adds a functionality for Fedora and for hopefully for another distributions.
qmake PREFIX=<folder_for_binary> and PREFIX_USR_SHARE=<folder_for_usr_share>
This can be used for INSTALL_ROOT case.

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -14,8 +14,16 @@ use_webengine {
     DEFINES += USE_WEBENGINE
 }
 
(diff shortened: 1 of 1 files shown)


trollixx:
Should be resolved by 991d3f80e6fb454df66d887d8419a0dd9f7eea48 in a more standard way.

PR #284, opened 2015-02-23
Skip prettyProductName and currentCpuArchitecture if not on QT5.4+

k3rni:
This makes Zeal compile with no errors on QT 5.3.2, as shipped with Debian Jessie.


trollixx:
Fixed in a little bit more verbose way in 281c550056edbbea8f206ab7620f92fa203e6440. Thanks for brining my attention to the failing compilation.

PR #297, opened 2015-03-26
Fixed unwanted reset of overall progress bar and download button text wh...

mNisblee:
...en downloading multiple docsets.

* Added a simple if(replies.isEmpty()) to prevent progress bar and button text reset when downloads are pending.
* Removed a redundant setText() from download button slot. (Already called via: on_downloadDocsetButton_clicked()->downloadDashDocset()->startDownload())


trollixx:
Included in 5b3106 and 3d1dca. Thanks!

PR #351, opened 2015-06-18
bigger settings dialog

mathieujobin:
Hello,

thank you for making this app. one thing I don't like when using Windows, is that all settings dialog are always too small. especially unresizable list views.

thanks to you, you've written this app in Qt ! Hurray

so here, i am doing 3 things.

* adding maximize button on settings dialog
* set first tab automatically on settingsdialog controller. otherwise, that last tab remains select even after app restart and the downloadable docsets list isn't loaded properly.
* I am adding a QGroupBox around the top part for installed docsets.

I tried to make the two groupbox into a vertical layout that includes a splitter, but I could not get QtCreator to do that for me.

let me know if you are interesting merging, I will solve conflit.
I hope it's not in the XML, but I have strong feeling it must be, isn't it ?

### DIFF
--- a/src/ui/settingsdialog.cpp
+++ b/src/ui/settingsdialog.cpp
@@ -95,6 +95,7 @@ SettingsDialog::SettingsDialog(Core::Application *app, ListModel *listModel, QWi
             this, &SettingsDialog::extractionProgress);
 
     loadSettings();
+    ui->tabWidget->setCurrentIndex(0);
 }
 
 SettingsDialog::~SettingsDialog()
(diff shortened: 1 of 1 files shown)


mathieujobin:
I took out my 3rd commit which had changes to the .ui file.
since you also added the installed docsets group

cheers,

trollixx:
Hi, thank you!

1. I really don't like the maximisation button in dialogue windows, it is possible to resize the window to your needs, but maximising it feels like an overkill to me. I understand that the main issue is experience with docset lists. I am going to move (actually reimplement) docset management into a separate "Docset Manager", and it will be part of 0.2 release cycle (no ETA yet). 

2. Setting the first tab as current should not be required unless you save `settingsdialog.ui` with another tab selected. Please, double check this issue.

mathieujobin:
I see. Thanks for your detailed answer.

PR #365, opened 2015-07-02
fix: skip cache update if downloaded docset list is corrupted

pbera:
- bug
 - cached docset list may become corrupted
 - download feature may become inaccessible
 - annoying message box is displayed every time
- quick fix
 - delete the corrupted docset list from cache


trollixx:
Thanks for contributing. I am working on a reliable way to retrieve the docset list, so many different issues will be addressed at once. I am closing this, as it's more like a workaround, rather than the fix for the actual problem.

PR #371, opened 2015-07-12
Add a command-line option to list installed docsets and exit

### LINKED ISSUE #368: Get installed docsets from the command line
Id would be nice to get a list of installed docsets from zeal via the command line.
For example:
```bash
$ zeal --installed-docsets
python2.7 python3 html
```  
This can be useful for IDE/texteditor plugins to limit a requested search to a specific language and the installed docsets

k3rni:
Fixes #368

### DIFF
--- a/src/core/application.cpp
+++ b/src/core/application.cpp
@@ -7,7 +7,6 @@
 #include "ui/mainwindow.h"
 #include "util/version.h"
 
-#include <iostream>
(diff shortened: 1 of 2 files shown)


trollixx:
In #368 I explained why I don't want a CLI intended for non-human use. Closing this in favour of #400.

PR #373, opened 2015-07-13
Use libarchive from Homebrew of MacPorts on OSX

windoze:
Use libarchive from Homebrew of MacPorts on OSX as Apple doesn't ship headers.

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -67,6 +67,12 @@ win32 {
 macx {
     TARGET = Zeal
     ICON = resources/zeal.icns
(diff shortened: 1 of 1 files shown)


trollixx:
Hi,

Sorry, I won't merge this.

First of all, Zeal does not support OS X officially and it may break any time. I can't test anything OS X related, and those willing to try Zeal there are on their own. There's [Dash](https://kapeli.com/dash) for OS X, which works and looks great.

Another issue is that Homebrew is just one of the ways to obtain the library Zeal needs, so I don't think it's reasonable to hard-code specific paths.

PR #381, opened 2015-07-20
fix missing include for qt5.x

sjas:


trollixx:
Thanks, I fixed the problem in 0.1 branch (79fe23b8baca4179ea18e673225ee104a2ffded5), and then it's been merged in master.

PR #422, opened 2015-11-11
support delay query after typing

comicfans:
zeal makes query everytime a char typed, this makes lots query before complete word
typed. this model not work very well with lots docset  . this commit introduce a delay time
which user can adjust after char typing ,so most unexpected query filtered out. user can
still press return to query immediately. to keep zeal default behavior as previously, this
setting default set to zero (disabled)

### DIFF
--- a/src/core/settings.cpp
+++ b/src/core/settings.cpp
@@ -41,6 +41,7 @@ namespace {
 const char GroupBrowser[] = "browser";
 const char GroupDocsets[] = "docsets";
 const char GroupGlobalShortcuts[] = "global_shortcuts";
+const char GroupQueryBehavior[] = "query_behavior";
 const char GroupInternal[] = "internal";
 const char GroupState[] = "state";
 const char GroupProxy[] = "proxy";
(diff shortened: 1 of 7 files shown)


trollixx:
Hi! It's a really hacky workaround for the freezing issue :) I need to move search queries to background threads, and perhaps utilise a centralised index like Dash. I really don't want to merge this, because it would only decrease motivation for a proper solution.

comicfans:
Image that paste a-very-long-word to search bar,  I'm not sure send queries multi times than final need can be faster(even with threaded query). besides that query result list refreshes frequency . delay query can avoid this and it didn't conflict with centralised index or threaded query, right?  anyway I've read that per-docset per-thread model is working on so hopes to try that and it can provide better response :）

trollixx:
Pasting a word fires only one text change signal, so one query only, typing is the problem. The idea is to cancel all pending background threads when search query is updated. I'd like to have instant updates as you type, it often allows to stop typing early.

comicfans:
my mistake that I didn't confirm the paste case... so forgot this .

Issue #436, opened 2015-11-25
Zeal crashes when removing active docset


lusitania:
Specifics:
 1. active docset: Puppet (some query result is displayed)
 1. click remove
 1. crash

PR #443, opened 2015-12-05
Fix issue 436

Serge45:
Fix issue #436 by
 - Implement removeRows function for SearchModel
 - Remove all items in SearchModel for active tab after removing corresponding docset.

### DIFF
--- a/src/registry/searchmodel.cpp
+++ b/src/registry/searchmodel.cpp
@@ -97,6 +97,36 @@ int SearchModel::columnCount(const QModelIndex &parent) const
     return 2;
 }
 
+bool SearchModel::removeRows(int row, int count, const QModelIndex &parent)
+{
+    if (row + count <= m_dataList.size() && !parent.isValid()) {
+        beginRemoveRows(parent, row, row + count - 1);
+        while (count) {
+            m_dataList.removeAt(row);
+            --count;
+        }
+        endRemoveRows();
+        return true;
+    }
+    return false;
+}
+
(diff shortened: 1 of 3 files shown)


drognanar:
When the removed docset is a part of search results zeal still crashes.

jkozera:
It still crashed for me, I suspect because of missing `beginRemoveRows` and `endRemoveRows` calls. I've created a new PR (#457) to address this. (It now seems to work right, no crashes, with the addition of a131234beaa61e3bc1c30b74567378fc7b998d6c.)

trollixx:
Thank you for looking in the issue.

`searchState->sectionsList` can be just reset with `setResults();` instead of removing items row by row.

I don't really like the idea of iterating over the whole `searchState->zealSearch` to remove docset-specific items, it might take a while when iterating over decent number of heavy docsets. But it's still better than crashing.

First of all, please rebase the change on top of `master` branch, and squash all commits. It would also make sense to incorporate changes from @jkozera's pull request, as I don't want to merge a broken commit. I hope Jerzy would not object :)

Also, commit message should be something like `registry, ui: Fix crash on docset removal (fixes #436)`.

Serge45:
Hi, I've updated this PR based on @jkozera 's PR(#457) and squashed all commits, but after rebase, @jkozera 's commit message disappears. Is there any way to keep it?

jkozera:
Don't worry, I don't mind losing that commit message with the rebase, so feel free to merge it as is. :)

PR #444, opened 2015-12-09
Index creation query failed for ZDash docsets

joaomsa:


trollixx:
Great catch! Could you please update the commit message to `registry: Fix index creation for ZDash docsets`?

joaomsa:
Sorry, didn't notice Zeal used imperative style commit messages. Fixed :)

trollixx:
Thanks!

Issue #354, opened 2015-06-19
Dedicated docset management UI


trollixx:
Docsets should be managed via a dedicated UI, not in preferences.

PR #456, opened 2016-01-04
Adds an input bar for filtering available docsets

brunoro:
Makes browsing through the list of downloadable docsets easier.
filter_docsets
filter_docsets_php

### DIFF
--- a/src/ui/forms/settingsdialog.ui
+++ b/src/ui/forms/settingsdialog.ui
@@ -466,6 +466,13 @@
            </widget>
           </item>
           <item>
+           <widget class="QLineEdit" name="docsetFilterInput">
+            <property name="placeholderText">
+             <string>filter docsets</string>
+            </property>
+           </widget>
+          </item>
+          <item>
            <layout class="QHBoxLayout" name="horizontalLayout">
             <item>
              <widget class="QLabel" name="label">
--- a/src/ui/settingsdialog.cpp
+++ b/src/ui/settingsdialog.cpp
@@ -111,6 +111,8 @@ SettingsDialog::SettingsDialog(Core::Application *app, ListModel *listModel, QWi
             this, &SettingsDialog::updateAllDocsets);
     connect(ui->removeDocsetsButton, &QPushButton::clicked,
             this, &SettingsDialog::removeSelectedDocsets);
+    connect(ui->docsetFilterInput, &QLineEdit::textEdited,
+            this, &SettingsDialog::updateDocsetFilter);
 
     ui->availableDocsetList->setItemDelegate(new ProgressItemDelegate(this));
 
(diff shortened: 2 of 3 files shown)


jkozera:
:+1: Thanks! Looks good to me, let's hope @trollixx gets some spare time for reviewing it too. :)

joaomsa:
:+1: works great for me too

trollixx:
It's a nice feature. The problem here is the unhiding of already installed docsets. Here's a an improved version of `SettingsDialog::updateDocsetFilter()` method, that fixes the problem and eliminates extra loop over the whole list (in `QListWidget::findItems()`):

```c++
void SettingsDialog::updateDocsetFilter(const QString &filterString)
{
    const bool doSearch = !filterString.simplified().isEmpty();

    for (int i = 0; i < ui->availableDocsetList->count(); ++i) {
        QListWidgetItem *item = ui->availableDocsetList->item(i);

        // Skip installed docsets
        if (m_docsetRegistry->contains(item->data(ListModel::DocsetNameRole).toString()))
            continue;

        item->setHidden(doSearch && !item->text().contains(filterString, Qt::CaseInsensitive));
    }
}
```

Please also update the commit message to something like `ui: Implement filtering of available docsets in Settings`. 

I'll merge as soon as soon as it's ready. Thanks!

brunoro:
Why would you like to skip the installed docsets? I'd see it as a valid use case to use the filter bar to find a docset I want to uninstall.

Also, is there any advantage of doing multiple `contains` calls on QString instead of using the `findItems` method of QList, or is it just a matter of style?

trollixx:
> Why would you like to skip the installed docsets? I'd see it as a valid use case to use the filter bar to find a docset I want to uninstall.

The installed docsets are already visible in the list above. The `Available Docsets` list is only for docsets that can be installed.

> Also, is there any advantage of doing multiple `contains` calls on QString instead of using the `findItems` method of QList, or is it just a matter of style?

If you check the source if [`QListWidget::findItems()`](http://code.woboq.org/qt5/qtbase/src/widgets/itemviews/qlistwidget.cpp.html#_ZNK11QListWidget9findItemsERK7QString6QFlagsIN2Qt9MatchFlagEE), you'll see that [`QAbstractItemModel::match()`](http://code.woboq.org/qt5/qtbase/src/corelib/itemmodels/qabstractitemmodel.cpp.html#_ZNK18QAbstractItemModel5matchERK11QModelIndexiRK8QVarianti6QFlagsIN2Qt9MatchFlagEE) it calls does even more complicated stuff inside. And [then](http://code.woboq.org/qt5/qtbase/src/widgets/itemviews/qlistwidget.cpp.html#1728) there's an additional loop over results returned by `QAbstractItemModel::match()`.

brunoro:
I see. I'll make the changes later and update the PR.

drognanar:
Call to `QListWidget::findItems` is already used in `SettingsDialog::findDocsetListItem`. Personally, to keep consistency I would make both methods manually iterate over items or call `
QListWidget::findItems()`.

drognanar:
Also it seems that more and more places are responsible for managing the visibility state of available docsets. As usual this causes bugs. One example is if the docset directory is changed. A docset is then removed from the installed docsets list but still hidden in available docsets list.

To resolve the issue `updateDocsetFilter` should also be called once the available list gets initially populated and also whenever registry emits an add/remove event.

trollixx:
@drognanar that's a valid point, see 32af98c00e6b3bd73671afc02e9c4540a0d6ec88.

trollixx:
Regarding the second observation, please create a ticket, I'll fix that once this change is merged.

In fact I believe the most of the SettingsDialog code should go away (with #354) or be rewritten. For instance, download/unpack progress indication is also broken, and many other things.

brunoro:
rebased + updated `SettingsDialog::updateDocsetFilter`

trollixx:
Merged, thanks!

brunoro:
(:

PR #457, opened 2016-01-04
Improved fix for issue #436

jkozera:
@Serge45's fix for #436 at the PR #443 doesn't call `beginRemoveRows` and `endRemoveRows` when necessary. This PR attempts to improve over that PR by calling the appropriate methods.

I'm not sure whether my `setUpdatesEnabled(false)` and `setUpdatesEnabled(true)` calls make sense, but I couldn't find a better way to avoid refreshing the whole list on every search result row's removal. (The range removed there can have gaps, so you cannot in general reduce it to a single {begin,end}RemoveRows pair.)


Serge45:
Thanks for your improvement :)

trollixx:
Closing this since #443 incorporated the changes.

Issue #410, opened 2015-10-10
Suggestion to make the "see also" area resizable


Miguellissimo:
Hi, I have a suggestion for a small improvement.  When opening a new page the area displaying the list of installed docsets gets cut in half and a "see also" area appears. It would be nice if a user could resize both areas by dragging the border vertically.

trollixx:
Fixed in 522e5148c9c03b2c0e1b3db3365f1880f41cdcc6.

Issue #262, opened 2015-01-26
Querying Results show imcompleted


showbye:
Search "printf" for example, the returning items show partly like "print" and 'f' is invisible,just in windows version.
![default](https://cloud.githubusercontent.com/assets/1239362/5900373/42b94bac-a5a3-11e4-9c1c-bf1a0133ee25.PNG)
It seems happened in every windows realease version.

trollixx:
Does it still occur with more recent Zeal versions? I cannot reproduce the bug.

showbye:
Problem still in the latest build .
Changing '2' to a big number in file searchitemdelegate.cpp 、line 44 seems better.

line 44:     rect.adjust(margin, 0, 2, 0); // +2px for bold text

trollixx:
I think this is fixed by 6c2401da6157cdadf9a2844a25ee355c98b13f22. Cannot reproduce anymore.

PR #459, opened 2016-01-07
Ui

drognanar:
Updates the UI.

Uses Ionicons for button icons.
Adds a resizable splitter for the see also panel (issue #410).
Adds a new tab button.
Adds tooltips to buttons.
Fix quirks with the toc delegate under windows (issue #262).
Makes UI rendering more consistent by separating state changes and UI changes.


trollixx:
These all are great improvements. Unfortunately, I really don't know how to approach this PR. Stupid GitHub, it would be so great to have each commit as a separate pull-request. Should I start commenting the code, it might be quite annoying to search what commit a specific comment is related to.

There are several ways we can proceed:
* You split this PR on per commit basis and we deal with them one by one.
* We handle all of them at once here, if you are OK with a good amount of interactive rebasing.
* I could amend everything as needed and merge independently. Of course, all commits would keep you as an author.

You choose :)

drognanar:
I managed to split the PR into three separate PRs as the commits were rather independent.
For the remaining work required (includng the perf PR) I'd rather do any changes necessary as a series of extra commits/rebases.

mrhota:
Is there a way anyone else could we give a hand so this can be moved along?

mrhota:
@trollixx what's the status? anything I can do to help? Some great performance improvements are waiting on this.

trollixx:
Changes, that were split from this PR, have been already merged. What's left is all mixed in one commit, and cannot be merged as-is. Regarding the second PR with performance fixes, I am looking into it, and hopefully get it merged soon.

drognanar:
Another update
The PR is now smaller in size and contains the first part of UI changes

trollixx:
After looking more into this, I doesn't look like the right approach.

On Linux Zeal relies on FreeDesktop icon theme, otherwise it wouldn't look consistent with other apps.

As for Windows, we should bundle some icons, and then set icon theme [path](http://doc.qt.io/qt-5/qicon.html#setThemeSearchPaths) and [name](http://doc.qt.io/qt-5/qicon.html#setThemeName), so those icons would be picked up automatically. Perhaps, there's a way to retrieve system icons used by other Windows applications, but I am unaware how.

I am closing this PR, but still thank you for contributing!

Commit b87890c275430fdcf5d17199dfbf9ef9836492da (`ui: Convert go to url button into a ToolButton `) has been incorporated as one of the recent UI refactoring changes.

PR #460, opened 2016-01-07
Improve search performance (caching, cancellation, parallel search)

drognanar:
Improves search performance. Returns results **without any observable delay** when searching through large docsets (including .NET and QT).

Cancels previous incomplete searches.
Defers browser url loading.
Adds caching of previous searches. Filters through the existing result list to avoid searching through the entire docset.
Adds parallel searching of docsets.

Depends on #459 in order to correctly display ToC after ESC is pressed.

### DIFF
--- a/src/registry/cachingsearchstrategy.cpp
+++ b/src/registry/cachingsearchstrategy.cpp
@@ -1,0 +1,98 @@
+/****************************************************************************
+**
+** Copyright (C) 2015 Oleg Shparber
+** Copyright (C) 2013-2014 Jerzy Kozera
+** Contact: http://zealdocs.org/contact.html
+**
+** This file is part of Zeal.
+**
+** Zeal is free software: you can redistribute it and/or modify
+** it under the terms of the GNU General Public License as published by
+** the Free Software Foundation, either version 3 of the License, or
+** (at your option) any later version.
+**
+** Zeal is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
+** along with Zeal. If not, see <http://www.gnu.org/licenses/>.
+**
+****************************************************************************/
+
+#include "cachingsearchstrategy.h"
+
+#include "docset.h"
+#include "searchquery.h"
+#include "searchresult.h"
+
+using namespace Zeal;
+
+const int CacheSize = 10;
+
+CachingSearchStrategy::CachingSearchStrategy(std::unique_ptr<DocsetSearchStrategy> strategy)
+    : m_search(std::move(strategy))
+    , m_cache(CacheSize)
+{
+}
+
+QList<SearchResult> CachingSearchStrategy::search(const QString &query, const CancellationToken &token)
+{
+    QString candidate = getCacheEntry(query);
+    if (!candidate.isEmpty()) {
+        return searchWithCache(query, candidate);
+    } else {
+        return searchWithoutCache(query, token);
+    }
+}
+
+bool CachingSearchStrategy::validResult(
+        const QString &query, const SearchResult &previousResult,
+        SearchResult &result) const
+{
+    return m_search->validResult(query, previousResult, result);
+}
+
+QString CachingSearchStrategy::getCacheEntry(const QString &query) const
+{
+    const SearchQuery searchQuery = SearchQuery::fromString(query);
+
+    for (int i = searchQuery.query().size(); i > 0; i--) {
+        QString candidate = searchQuery.query().mid(0, i);
+
+        /// Use the cache only if the cache entry contains all results.
+        /// Also handle cases where prefix is a docset query `std:`.
+        if (m_cache.contains(candidate)
+                && m_cache[candidate]->size() < Docset::MaxDocsetResultsCount
+                && SearchQuery::fromString(candidate).query().size() == i)
+            return candidate;
+    }
+    return QString();
+}
+
+QList<SearchResult> CachingSearchStrategy::searchWithCache(const QString &query, const QString &prefix)
+{
+    QList<SearchResult> *prefixResults = m_cache[prefix];
+    QList<SearchResult> results;
+    QListIterator<SearchResult> prefixResultsIterator(*prefixResults);
+    while (prefixResultsIterator.hasNext()) {
+        SearchResult previousResult = prefixResultsIterator.next();
+        SearchResult result;
+        if (validResult(query, previousResult, result))
+            results.append(result);
+    }
+
+    m_cache.insert(query, new QList<SearchResult>(results));
+    return results;
+}
+
+QList<SearchResult> CachingSearchStrategy::searchWithoutCache(const QString &query, CancellationToken token)
+{
+    QList<SearchResult> results = m_search->search(query, token);
+
+    /// Only cache the results if they are not partial.
+    if (!token.isCanceled())
+        m_cache.insert(query, new QList<SearchResult>(results));
+    return results;
+}
--- a/src/registry/cachingsearchstrategy.h
+++ b/src/registry/cachingsearchstrategy.h
@@ -1,0 +1,63 @@
+/****************************************************************************
+**
+** Copyright (C) 2015 Oleg Shparber
+** Copyright (C) 2013-2014 Jerzy Kozera
+** Contact: http://zealdocs.org/contact.html
+**
+** This file is part of Zeal.
+**
+** Zeal is free software: you can redistribute it and/or modify
+** it under the terms of the GNU General Public License as published by
+** the Free Software Foundation, either version 3 of the License, or
+** (at your option) any later version.
+**
+** Zeal is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
+** along with Zeal. If not, see <http://www.gnu.org/licenses/>.
+**
+****************************************************************************/
+
+#ifndef CACHINGSEARCHSTRATEGY_H
+#define CACHINGSEARCHSTRATEGY_H
+
+#include "cancellationtoken.h"
+#include "docsetsearchstrategy.h"
+
+#include <QCache>
+#include <QList>
+#include <QString>
+
+#include <memory>
+
+namespace Zeal {
+
+struct SearchResult;
+
+/// A search strategy that uses a cache of previous searches.
+/// If a search was made for a prefix then then all results
+/// must appear in the cache. In this case instead of searching
+/// entire docset only cache is searched.
+class CachingSearchStrategy : public DocsetSearchStrategy
+{
+public:
+    CachingSearchStrategy(std::unique_ptr<DocsetSearchStrategy> strategy);
+    QList<SearchResult> search(const QString &query, const CancellationToken &token) override;
+    bool validResult(const QString &query, const SearchResult &previousResult,
+                     SearchResult &result) const override;
+
+private:
+    QString getCacheEntry(const QString &query) const;
+    QList<SearchResult> searchWithCache(const QString &query, const QString &prefix);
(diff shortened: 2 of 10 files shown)


trollixx:
This is an awesome work. Thank you!

I'll be going over and cherry-picking commits individually shortly. I wish GitHub could create series of PRs like Gerrit does.

src/registry/docset.cpp
```diff
@@ -412,6 +472,7 @@ void Docset::countSymbols()
         const QString symbolType = parseSymbolType(symbolTypeStr);
         m_symbolStrings.insertMulti(symbolType, symbolTypeStr);
         m_symbolCounts[symbolType] += query.value(1).toInt();
+        m_symbolsTotal += query.value(1).toInt();
```
trollixx: I don't see this being used anywhere...

src/ui/mainwindow.cpp
```diff
@@ -148,7 +149,7 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     connect(ui->actionCheckForUpdate, &QAction::triggered,
             m_application, &Core::Application::checkForUpdate);
     connect(ui->actionAboutZeal, &QAction::triggered, [this]() {
-        QScopedPointer<AboutDialog> dialog(new AboutDialog(this));
+        std::unique_ptr<AboutDialog> dialog(new AboutDialog(this));
```
trollixx: I am really curious, apart from the support for move semantics that we don't use, what's the actual benefit from switching from `QScopedPointer` to `std::unique_ptr`?

src/ui/mainwindow.h
```diff
@@ -134,17 +140,22 @@ private slots:
...
+    std::unique_ptr<Zeal::SettingsDialog> m_settingsDialog;
+
+    std::unique_ptr<QMenu> m_backMenu;
+    std::unique_ptr<QMenu> m_forwardMenu;
 
-    QMenu *m_backMenu = nullptr;
-    QMenu *m_forwardMenu = nullptr;
+    Zeal::CancellationToken m_cancelSearch;
 
+    std::unique_ptr<QShortcut> m_focusSearch;
```
trollixx: I don't see how keeping this is useful.

src/ui/mainwindow.cpp
```diff
@@ -99,9 +101,9 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     // initialise ui
     ui->setupUi(this);
 
-    QShortcut *focusSearch = new QShortcut(QKeySequence(QStringLiteral("Ctrl+K")), this);
-    focusSearch->setContext(Qt::ApplicationShortcut);
-    connect(focusSearch, &QShortcut::activated,
+    m_focusSearch = std::unique_ptr<QShortcut>(new QShortcut(QKeySequence(QStringLiteral("Ctrl+K")), this));
```
trollixx: `QShortcut` is `QObject`, it is deleted when by its parent. Why `std::unique_ptr`?

src/ui/mainwindow.h
```diff
@@ -134,17 +140,22 @@ private slots:
...
     SearchState *m_searchState = nullptr;
-    Zeal::NetworkAccessManager *m_zealNetworkManager = nullptr;
 
-    Ui::MainWindow *ui = nullptr;
+    std::unique_ptr<Ui::MainWindow> ui;
     Zeal::Core::Application *m_application = nullptr;
     Zeal::Core::Settings *m_settings = nullptr;
-    Zeal::ListModel *m_zealListModel = nullptr;
-    Zeal::SettingsDialog *m_settingsDialog = nullptr;
+    std::unique_ptr<Zeal::NetworkAccessManager> m_zealNetworkManager;
```
trollixx: Same as with `QShortcut`, all these are `QObject`s with set parent, there's no need in any additional memory management.

src/registry/docsetsearchstrategy.h
```diff
@@ -0,0 +1,50 @@
...
+
+/// This class contains search related methods for docsets.
+class DocsetSearchStrategy
+{
+public:
+    DocsetSearchStrategy();
+    virtual QList<SearchResult> search(const QString &query, CancellationToken token) = 0;
+
+    /// Used to filter out cached results.
+    virtual bool validResult(const QString &query, SearchResult previousResult,
```
trollixx: Why make a copy of `previousResult`?

src/registry/docsetsearchstrategy.h
```diff
@@ -0,0 +1,50 @@
...
+namespace Zeal {
+
+class SearchResult;
+
+/// This class contains search related methods for docsets.
+class DocsetSearchStrategy
+{
+public:
+    DocsetSearchStrategy();
+    virtual QList<SearchResult> search(const QString &query, CancellationToken token) = 0;
```
trollixx: Why make a copy of `token`?

src/registry/docsetregistry.cpp
```diff
@@ -121,25 +125,32 @@ void DocsetRegistry::_addDocset(const QString &path)
...
-
-    for (Docset *docset : docsets())
-        m_queryResults << docset->search(query);
+    finalResult += partial;
+}
 
-    std::sort(m_queryResults.begin(), m_queryResults.end());
+void DocsetRegistry::_runQueryAsync(const QString &query, const CancellationToken token)
+{
+    QFuture<QList<SearchResult> > queryResultsFuture = QtConcurrent::mappedReduced(
```
trollixx: I really like this. I think we could utilise `QFutureWatcher` here, and get rid of the `QMetaObject::invokeMethod()` and `m_thread`, right?

Also, no need in the space between `>>`, C++11 is required to build the project.

src/registry/cancellationtoken.h
```diff
@@ -0,0 +1,48 @@
...
+#ifndef CANCELLATIONTOKEN_H
+#define CANCELLATIONTOKEN_H
+
+#include <QSharedPointer>
+
+namespace Zeal {
+
+/// Token that stores whether cancel was called on it.
+/// In async code can be used to check if another thread called cancel.
+struct CancellationToken
```
trollixx: As we briefly discussed, it feels like we could use `QFutureWatcher::cancel()` instead of a custom solution. Have you got a chance to look into it?

drognanar: I finally had the chance. `QFutureWatcher::cancel(); QFutureWatcher::waitForFinished()` waits until all existing threads finish and does not start new ones. Thus, worst case, `cancel` would take as much time as it takes to search the largest docset.
On the other hand `CancellationToken` allows to stop a search thread at any point during its execution.

src/registry/docsetregistry.cpp
drognanar: It definitely is possible to get rid of `invokeMethod`, but I'd rather do that in a separate PR.
Also, I was used to writing `> >` and only realized that this is not necessary after pushing this PR.

drognanar:
Made changes to the PR. I rolled back changes I made with regards to pointers.

src/registry/cachingsearchstrategy.cpp
```diff
@@ -0,0 +1,96 @@
...
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
+** along with Zeal. If not, see <http://www.gnu.org/licenses/>.
+**
+****************************************************************************/
+
+#include "cachingsearchstrategy.h"
+#include "docset.h"
```
trollixx: Please sort alphabetically, adding an empty line after `cachingsearchstrategy.h`.

src/registry/cachingsearchstrategy.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+#include "searchresult.h"
+#include "searchquery.h"
+
+using namespace Zeal;
+
+CachingSearchStrategy::CachingSearchStrategy(std::unique_ptr<DocsetSearchStrategy> strategy)
+    : m_search(std::move(strategy)),
```
trollixx: Zeal is currently uses other style, but I do think to switch over. Anyway, in this case comma should be on the next line.

src/registry/cachingsearchstrategy.h
```diff
@@ -0,0 +1,65 @@
...
+    QList<SearchResult> search(const QString &query, CancellationToken &token) override;
+    bool validResult(const QString &query, const SearchResult &previousResult,
+                     SearchResult &result)override;
+
+private:
+    QString getCacheEntry(const QString &query) const;
+    QList<SearchResult> searchWithCache(const QString &query, const QString &prefix);
+    QList<SearchResult> searchWithoutCache(const QString &query, CancellationToken token);
+
+    const static int CacheSize = 10;
```
trollixx: I think this should live in the .cpp file scope.

src/registry/cancellationtoken.cpp
```diff
@@ -0,0 +1,41 @@
...
+using namespace Zeal;
+
+CancellationToken::CancellationToken()
+{
+    m_cancelled = QSharedPointer<bool>(new bool(false));
+}
+
+void CancellationToken::cancel()
+{
+    *m_cancelled = true;
```
trollixx: This is not an atomic operation, but I am not sure if it can cause any problems...

src/registry/docsetregistry.cpp
```diff
@@ -120,25 +124,32 @@ void DocsetRegistry::_addDocset(const QString &path)
...
 
-void DocsetRegistry::search(const QString &query)
+void DocsetRegistry::search(const QString &query, CancellationToken token)
 {
-    // Only invalidate queries
-    if (query.isEmpty())
-        return;
-
-    QMetaObject::invokeMethod(this, "_runQuery", Qt::QueuedConnection, Q_ARG(QString, query));
+    qRegisterMetaType<CancellationToken>("CancellationToken");
```
trollixx: Why register both?

src/registry/docsetregistry.cpp
```diff
@@ -120,25 +124,32 @@ void DocsetRegistry::_addDocset(const QString &path)
...
+    finalResult += partial;
+}
 
-    std::sort(m_queryResults.begin(), m_queryResults.end());
+void DocsetRegistry::_runQueryAsync(const QString &query, const CancellationToken token)
+{
+    QFuture<QList<SearchResult> > queryResultsFuture = QtConcurrent::mappedReduced(
+                docsets(),
+                std::bind(&Docset::search, std::placeholders::_1, query, token),
+                &MergeQueryResults);
```
trollixx: I'd prefer a small lambda instead of `MergeQueryResults`.

src/registry/cancellationtoken.cpp
drognanar: There are no write races since mainwindow is the only place where the token gets cancelled.
And even if there would be many threads cancelling all set the value to true.
Finally, the shared pointer is not reassigned and cannot be disposed since at least one instance of CancellationToken exists (the one that is calling the cancel method).

src/registry/docsetregistry.cpp
drognanar: I remember that I had issues when registering the meta type here earlier.
However, I no longer recall the exact problem and after removing one of the calls did not break zeal.

src/registry/docsetregistry.cpp
drognanar: I tried but the program did not compile when lambda was passed in.

drognanar:
Another update

mrhota:
@drognanar Any updates here? I can rebase this if you don't have the time.

mrhota:
@trollixx @drognanar what needs to happen to move this along? Fuzzy searching is waiting on this.

trollixx:
I've been going over this multiple times, and the implementation details seem very counterintuitive to me. Exposing docset internals outside of `Docset` class is one of such things. It's quite opposite from what I was trying to achieve by encapsulating everything in one class.

I am really not comfortable with merging it right away, as I am not completely sold on search strategies idea. Chaining multiple is not a really flexible approach. For example, it leads to  iterating over results directly, while there's no way to apply custom search logic. Also the order in which strategies can be chained is also fixed, what I'd consider a design flaw.

There are a decent amount of unrelated changes here, like replacing `QScopedPointer` with `std::unique_ptr`, removing `Docset::path()`, etc.

I'd prefer to hold this off a bit longer, and see how's other changes from this PR will do in the next release.

trollixx:
This is a very old and abandoned PR, let's put it to rest.

PR #472, opened 2016-01-16
ui: Fix SearchItemDelegate display on windows

drognanar:
Ensures that the highlight for SearchItemDelegate on Windows is
displayed correctly.


trollixx:
I ran this with `-style windows`, and here what I saw:

drognanar:
It seems that the old windows style did not highlight the icon, thus making the highlight look weird.
I updated the PR and the new implementation should work in the general case.

trollixx:
It looks like with 601da85cd8e2fd0da794ee69a141d2482e1db075 the problem is gone now. Could you please verify that as well? I've only tested on Windows 10 and on Linux with `-style windows`.

trollixx:
Just an update to this. I've spent a lot of time trying to make bold highlight work properly. It seems to be almost impossible to implement and cover all corner cases. So, instead of messing with a complex logic, I've decided to go with a highlighting via background.

WIP preview:


Unfortunately, during the testing on Windows 10, I have discovered [a bug](https://bugreports.qt.io/browse/QTBUG-50976) in Qt, that messes up font sizes in `QListView` or `QTreeView`, when Qt Style Sheets are in use. So, now I am removing the style sheets for splitters and frames.

Long story short: I am closing this PR, because I will be pushing a brand new highlighting implementation shortly. It's in `feature/searchitemdelegate` branch at the moment.

I'll also start reviewing #459 and #460 (atomic commits would be great), once I am done with the highlighting.

Issue #447, opened 2015-12-10
book-like chapter tree view?


str50:
Hi, 

What do you think of implementing a tree view in the main ui that looks more like chapters in a manual book? Many documentation already has a structure that is organized by package/library/guidelines etc. In other words, I am talking about TOC on the docset level.

I think this might be more useful than listing groups of symbols in their types. Seriously why would anyone care that there is a group of 1000+ functions in language/framework xx?

Thanks,

trollixx:
Unfortunately, there's no global TOC information in docsets. I thought about this feature in the past, but related to a support of CHM and PDF files.

@Kapeli have you considered adding a docset-wide TOC?

Kapeli:
A hierarchical docset index would have been nice, but unfortunately it's impossible to add right now as I don't see a way to do it which doesn't require re-indexing all of the docsets.

> Seriously why would anyone care that there is a group of 1000+ functions in language/framework xx?

I don't think anyone cares. The reason it's like this is that Dash was modelled as a dictionary and for a long time it didn't even allow browsing the docsets, it was just search-only. When I started allowing browsing it was too late to switch to a hierarchical index.

Issue #386, opened 2015-07-26
TOC Support


trollixx:
Zeal generates TOC ("See also") from index database, while it should be done by parsing DOM. Details: https://kapeli.com/docsets#tableofcontents.

PR #473, opened 2016-01-16
MainWindow refactor and resizable sections list

drognanar:


src/core/settings.cpp
```diff
@@ -117,9 +117,17 @@ void Settings::load()
     }
     m_settings->endGroup();
 
+
```
trollixx: whitespace

src/core/settings.cpp
```diff
@@ -117,9 +117,17 @@ void Settings::load()
...
+
     m_settings->beginGroup(GroupState);
     windowGeometry = m_settings->value(QStringLiteral("window_geometry")).toByteArray();
-    splitterGeometry = m_settings->value(QStringLiteral("splitter_geometry")).toByteArray();
+    verticalSplitterGeometry = m_settings->value(QStringLiteral("splitter_geometry")).toByteArray();
+
+    sectionsSplitterSizes = QList<int>();
+    QList<QVariant> splitterSizes = m_settings->value(QStringLiteral("sections_geometry")).toList();
+    for (int i = 0; i < splitterSizes.size(); i++)
```
trollixx: Please use range-based for here

src/core/settings.cpp
```diff
@@ -164,9 +172,14 @@ void Settings::save()
     m_settings->endGroup();
 #endif
 
+    QList<QVariant> splitterSizes;
+    for (int i = 0; i < sectionsSplitterSizes.size(); i++)
```
trollixx: Same as above

src/ui/mainwindow.cpp
```diff
@@ -104,9 +104,13 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
             ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
 
     restoreGeometry(m_settings->windowGeometry);
-    ui->splitter->restoreState(m_settings->splitterGeometry);
+    ui->splitter->restoreState(m_settings->verticalSplitterGeometry);
     connect(ui->splitter, &QSplitter::splitterMoved, [=](int, int) {
```
trollixx: Hm, actually we do store the geometry in `closeEvent()`, maybe it's excessive to update it on every move.

src/ui/mainwindow.cpp
```diff
@@ -104,9 +104,13 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
...
-        m_settings->splitterGeometry = ui->splitter->saveState();
+        m_settings->verticalSplitterGeometry = ui->splitter->saveState();
+    });
+    connect(ui->sectionsSplitter, &QSplitter::splitterMoved, [=](int, int) {
+        if (ui->sections->isVisible())
```
trollixx: How about storing the height of `sections` in `displaySections()` before hiding it?

src/ui/mainwindow.cpp
```diff
@@ -468,6 +461,38 @@ void MainWindow::createTab()
...
+
+    if (!searchState->searchQuery.isEmpty()) {
+        ui->treeView->setModel(searchState->zealSearch);
+        ui->treeView->setRootIsDecorated(false);
+        ui->treeView->reset();
+    } else {
+        ui->treeView->setModel(m_zealListModel);
+        ui->treeView->setColumnHidden(1, true);
+        ui->treeView->setRootIsDecorated(true);
+        ui->treeView->reset();
```
trollixx: Duplicated in the above block, so should be moved out.

src/ui/mainwindow.h
```diff
@@ -122,9 +122,12 @@ private slots:
 
 private:
     void displayViewActions();
+    void displayTreeView();
+    void displaySections();
     void setupSearchBoxCompletions();
     void reloadTabState();
     void displayTabs();
+    SearchState *currentSearchState();
```
trollixx: `const` is missing.

trollixx:
Mostly nitpicking, but I've spent so much time cleaning up the code base, so now I prefer to keep it pretty in the first place :)

In the commit description it should be either `MainWindow` or `main window`. And in the description either `Move logic...` or `This change moves...`. Commit comments should be in the imperative form.

Actually, the name `sections` does not carry much sense. It should be in fact `toc` and work differently (see #386). I have a renaming change, but now I'll wait until this change lands.

drognanar:
Updated PR. I'll rename the commits once everything else is fine.

Regarding the name for `sections` I am unsure about calling it `toc` either as the `treeView` acts more like a table of contents.

Also I am well aware that it should work differently. I made more progress on my master branch and one of commits adds support for .dashtoc files 969dfb4d9c8. However, I'm waiting for the current PRs to complete first.

trollixx:
Thanks for being patient. The resulting code looks good to me. Can you amend the relevant commits instead of pushing fixes on top of them?

Perhaps `pageToc` would work, but that's not important at the moment. Unfortunately, we don't have any real table of contents in docsets (#447).

drognanar:
I rebased the change. I normally do that after the fixes get accepted to make the review easier.

trollixx:
Yeah, it all looks good. Any idea why it has picked up my latest commit?

trollixx:
`git rebase master` fixes the problem in my local branch.

drognanar:
I rebased against upstream before and somehow it did not pick up one of the commits.
Fixed.

Issue #497, opened 2016-02-18
Segmentation fault (core dumped)


photous:
when running zeal in the terminal it says:
```
 Segmentation fault (core dumped)
```
I have tried to `apt-get purge zeal` then install it again but it did not solve my problem.

trollixx:
Try removing ~/.config/Zeal and ~/.local/share/Zeal.

photous:
Thank you , It's working now :)
Do the docsets cause the issue?

trollixx:
Yeah, apparently corrupted docset can cause a crash.

photous:
How can I know which docset is corrupted?

trollixx:
At the moment only by adding/removing docset folders one by one.

photous:
Thank you very much, I have successfully found the problem. It was an unfinished docset.


Example of a full docset's `Contents` directory:
````
ls Ruby_on_Rails_4.docset/Contents/Resources/
docSet.dsidx      docSet.dsidx-wal  docSet.toc         Documents  Nodes.xml
docSet.dsidx-shm  docSet.mom        docSet.tokencache  LICENSE    Tokens.xml
```
The unfinished docset's `Contents` directory looks like this:
```
ls Twisted.docset/Contents/Resources/
docSet.dsidx  LICENSE
````

I hope this will give a clear view of the issue and makes it easy to fix  :+1:

trollixx:
Thank you for the details. Hopefully fixed in f24039251c4d2e9e0d89ce19beb5a7763f26819b. Marking as a duplicate of #489, since it seems to be the same problem.

Issue #489, opened 2016-02-12
zeal crash on openning


agentready:
zeal crash every time I open it. In terminal I see this message --> 
```
Cannot determine index file for docset SQLite
QxtGlobalShortcut failed to register: Meta+Z
Segmentation fault (core dumped)
```
I've installed zeal from ppa and my os ubuntu 14.0.4

trollixx:
What's your docset list? Looks like one of the docsets is corrupted. Any were manually installed?

agentready:
Hi thanks for reply , here is my docset list --> http://pastebin.com/v1KRJfsk , I didn't manually installed anything. One thing I can remember while downloading a doc my laptop were hanged by another app and I did hard restarted the laptop.

trollixx:
A corrupted docset might cause the problem somehow sneaking through the checks.

You can try to remove SQLite docset folder, and see whether it is the source of the issue.

agentready:
Hi, Thanks again for the reply. Yes, after removing SQLite doc folder it only saying 
```
QxtGlobalShortcut failed to register: Meta+Z
Segmentation fault (core dumped)
```
After removing every docset folder, it is working except QxtGlobalShortcut.

trollixx:
My guess is that that not fully downloaded docset was an issue. Unfortunately, it's hard do tell what exactly caused an issue, unless you kept all folders, so we could check which one was corrupted one by one.

I've fixed the index file issue with SQLite docset. Requires a reload of the docset database, and reinstalling of the docset itself.

`QxtGlobalShortcut` issue is interesting, what your desktop environment?

PR #501, opened 2016-02-21
ui: Fix segfault for incomplete docsets

drognanar:
Possibly fixes issues #497 and #489


trollixx:
Thank you!

I've merged this manually because of the minor alterations that were not worth asking you to fix.

PR #507, opened 2016-03-04
ui: Prevent QWebEnginePage from stealing focus from the search input

sonald:
this fix the problem of losing focus instantly when typing

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -407,6 +407,9 @@ void MainWindow::queryCompleted()
     m_treeViewClicked = true;
     ui->treeView->setCurrentIndex(m_searchState->zealSearch->index(0, 0, QModelIndex()));
     ui->treeView->activated(ui->treeView->currentIndex());
+#if USE_WEBENGINE
+    ui->lineEdit->setFocus(Qt::MouseFocusReason);
+#endif
 }
 
 void MainWindow::goToTab(int index)


trollixx:
Hey, thanks for the contribution!

It seems that I cannot reproduce the problem. I think I did encounter this once, but that could be something different. Is the issue persistent for you, or do you have a reliable way to make it reproducible?

0x1997:
`git bisect` shows the transition to Qt WebEngine from Qt 5.6 in 99af6ea79f14ca00dd565fe46bd4ce30f86f37ab appears to be the culprit. Not sure why though.

drognanar:
The issue which I stumbled upon was that on WebEngine `QWebPage::load()` steals focus.
Thus disabling the web view for `load` or re focusing on line edit should prevent the focus from getting stolen.

sonald:
I can comfirm it happens only when using WebEngine

trollixx:
Thanks you all for the investigation!

Looks like the reason for our problem comes from [here](http://code.qt.io/cgit/qt/qtwebengine.git/tree/src/core/web_contents_adapter.cpp?id=0110b93af4ab61a5b86643c2be8c6f34648bbaaf#n489). I would really like to know what was the reasoning for forcing the focus...

trollixx:
I think I've found a little bit better solution. Checking if it works on Windows...

src/ui/mainwindow.cpp
```diff
@@ -407,6 +407,8 @@ void MainWindow::queryCompleted()
     m_treeViewClicked = true;
     ui->treeView->setCurrentIndex(m_searchState->zealSearch->index(0, 0, QModelIndex()));
     ui->treeView->activated(ui->treeView->currentIndex());
+    ui->lineEdit->setFocus();
```
trollixx: Use `Qt::MouseFocusReason` to avoid the need in `deselect()` below. Also this code is only relevant for Qt WebEngine, so a check for `USE_WEBENGINE` should be here. The following code works fine for me. Please check for yourself, and update the pull request.

```cpp
#ifdef USE_WEBENGINE
    ui->lineEdit->setFocus(Qt::MouseFocusReason);
#endif
```

trollixx:
@sonald Please check if my solution (in the comment to the code) works for you as well, and if everything is fine, update the commit. Also please set the commit message to `ui: Prevent QWebEnginePage from stealing focus from the search input`. Thanks!

sonald:
I've tested and it works fine.

trollixx:
There's a temporary file (`src/ui/.mainwindow.cpp.swp`) in the commit.

sonald:
oh, sorry for that mistake.

trollixx:
Merged, thanks!

PR #510, opened 2016-03-14
build for OSX using homebrew qt5 and libarchive

jrwren:


README.md
```diff
@@ -43,6 +43,12 @@ After installing Zeal, you need to download docsets. Go to *File->Options->Docse
 
 To compile Zeal run `qmake` and then `make`. Linux users can install Zeal with `make install` command.
 
+### Instructions for OSX
```
trollixx: I think a [wiki](https://github.com/zealdocs/zeal/wiki) page would be a more appropriate place for such information.

Putting this here would assume some kind of support for OS X, while there's none.

README.md
```diff
@@ -43,6 +43,12 @@ After installing Zeal, you need to download docsets. Go to *File->Options->Docse
...
+
+    brew install qt5 libarchive
+    /usr/local/opt/qt5/bin/qmake && make
+    open bin/Zeal.app
```
trollixx: `open` doesn't really look like a part of the build process...

src/src.pro
```diff
@@ -1,5 +1,8 @@
 TEMPLATE = app
 
+macx:INCLUDEPATH += /usr/local/opt/libarchive/include
```
trollixx: Wrong place :) Build settings for `libarchive` live in `src/core/core.pri`. And since `homebrew` is only one of the ways to obtain dependencies on OS X, I am not convinced these paths should hardcoded here... It's possible to specify them as `qmake` parameters: `qmake INCLUDEPATH+=... LIBS+=...`. So it might be better be in wiki as well.

trollixx:
I have created [a wiki page](https://github.com/zealdocs/zeal/wiki/Build-Instructions-for-OS-X) based on your PR. Please verify that it's correct (I am not an OS X user), and feel free to alter if needed.

Thanks for the contribution!

PR #517, opened 2016-03-20
Add Qt 5.6 support.

Red54:


trollixx:
This disables the ability to build Zeal with WebEngine for Qt below 5.6. The correct way would be making WebEngine default for Qt 5.6+.

trollixx:
Implemented in 99af6ea79f14ca00dd565fe46bd4ce30f86f37ab.

PR #529, opened 2016-04-14
Parse query given the list of known docset keywords

drognanar:
This is the first in order to implement docset keyword groups and custom docset keywords.

Adds a parameter to `SearchQuery::fromString` that has a list of available keywords.
Replaces all places where the query is constructed with the call to docsetregistry that is aware of currently available keywords.

In effect `prefix1:` will only be treated as a keyword prefix if it matches a valid keyword.
For example `std:` is not treated as a keyword since no docset has a name std.
In fact `cpp:` will not be treated as a keyword until the c++ docset is installed.

### DIFF
--- a/src/registry/docset.cpp
+++ b/src/registry/docset.cpp
@@ -241,11 +241,10 @@ const QMap<QString, QString> &Docset::symbols(const QString &symbolType) const
     return m_symbols[symbolType];
 }
 
-QList<SearchResult> Docset::search(const QString &query) const
+QList<SearchResult> Docset::search(const SearchQuery &searchQuery) const
 {
     QList<SearchResult> results;
 
-    const SearchQuery searchQuery = SearchQuery::fromString(query);
     const QString sanitizedQuery = searchQuery.sanitizedQuery();
 
     if (searchQuery.hasKeywords() && !searchQuery.hasKeywords(m_keywords))
--- a/src/registry/docsetregistry.cpp
+++ b/src/registry/docsetregistry.cpp
@@ -147,7 +147,7 @@ void DocsetRegistry::_runQuery(const QString &queryStr)
 
 SearchQuery DocsetRegistry::getSearchQuery(const QString &queryStr) const
 {
-    return SearchQuery::fromString(queryStr);
+    return SearchQuery::fromString(queryStr, completions());
 }
 
 QStringList DocsetRegistry::completions() const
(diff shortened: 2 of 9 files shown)


src/registry/docset.cpp
```diff
@@ -241,11 +241,10 @@ const QMap<QString, QString> &Docset::symbols(const QString &symbolType) const
     return m_symbols[symbolType];
 }
 
-QList<SearchResult> Docset::search(const QString &query) const
+QList<SearchResult> Docset::search(const SearchQuery &searchQuery) const
```
trollixx: The approach here is wrong, and this change doesn't make it better. `Docset::search()` doesn't need to know anything about keywords. We should move the check below to `DocsetRegistry`.

src/ui/mainwindow.cpp
```diff
@@ -534,15 +535,7 @@ void MainWindow::reloadTabState()
 // Sets up the search box autocompletions.
 void MainWindow::setupSearchBoxCompletions()
```
trollixx: Looks like a pointless method now.

src/ui/widgets/searchedit.cpp
```diff
@@ -154,7 +160,10 @@ QString SearchEdit::currentCompletion(const QString &text) const
 
 int SearchEdit::queryStart() const
 {
-    const Zeal::SearchQuery currentQuery = Zeal::SearchQuery::fromString(text());
+    const Zeal::SearchQuery currentQuery = m_registry != nullptr
```
trollixx: Is it possible not to depend on `DocsetRegistry` here? If not, we can make `m_registry` required, and then just complain and return 0 if it's not set.

src/registry/docsetregistry.cpp
```diff
@@ -143,6 +145,24 @@ void DocsetRegistry::_runQuery(const QString &query)
     emit queryCompleted(results);
 }
 
+SearchQuery DocsetRegistry::getSearchQuery(const QString &queryStr) const
+{
+    return SearchQuery::fromString(queryStr, completions());
+}
+
+QStringList DocsetRegistry::completions() const
```
trollixx: The completion list can be built in runtime as docsets are added and removed. Should be more efficient.

trollixx:
Closing this in favor of the eventually coming new docset library implementation.

PR #538, opened 2016-04-19
Added duplicate tab feature

### LINKED ISSUE #524: Duplicate tab feature
Should preserve current left pane state, search box input, web view scroll position.

Firefox uses <kbd>Ctrl+Alt+T</kbd> as a shortcut.

innermous:
Closes #524 
Maybe any remarks?

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -85,6 +85,41 @@ struct TabState
 #endif
     }
 
+    TabState(const TabState &rhs)
+        : searchQuery(rhs.searchQuery)
+        , selections(rhs.selections)
+        , expansions(rhs.expansions)
+        , searchScrollPosition(rhs.searchScrollPosition)
+        , tocScrollPosition(rhs.tocScrollPosition)
+        , webViewZoomFactor(rhs.webViewZoomFactor)
+    {
+        searchModel = new Zeal::SearchModel();
+        tocModel = new Zeal::SearchModel();
+
+        webPage = new QWebPage();
+#ifndef USE_WEBENGINE
+        webPage->setLinkDelegationPolicy(QWebPage::DelegateExternalLinks);
+        webPage->setNetworkAccessManager(Core::Application::instance()->networkManager());
+#endif
+
+        QByteArray historyArray = rhs.saveHistory();
+        restoreHistory(historyArray);
+    }
+
+    void restoreHistory(const QByteArray &array) const
+    {
+        QDataStream stream(array);
+        stream >> *webPage->history();
+    }
+
+    QByteArray saveHistory() const
+    {
+        QByteArray array;
+        QDataStream stream(&array, QIODevice::WriteOnly);
+        stream << *webPage->history();
+        return array;
+    }
+
     ~TabState()
     {
         delete searchModel;
@@ -106,6 +141,15 @@ struct TabState
         webPage->load(url);
 #else
         webPage->mainFrame()->load(url);
+#endif
+    }
+
+    QString title() const
+    {
+#ifdef USE_WEBENGINE
+        return webPage->title();
+#else
+        return webPage->mainFrame()->title();
 #endif
     }
 
@@ -154,6 +198,9 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     QShortcut *focusSearch = new QShortcut(QStringLiteral("Ctrl+K"), this);
     connect(focusSearch, &QShortcut::activated,
             ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
+
+    QShortcut *duplicate = new QShortcut(QStringLiteral("Ctrl+Alt+T"), this);
+    connect(duplicate, &QShortcut::activated, this, [this]() { duplicateTab(m_tabBar->currentIndex()); });
 
     restoreGeometry(m_settings->windowGeometry);
     ui->splitter->restoreState(m_settings->verticalSplitterGeometry);
@@ -467,6 +514,22 @@ void MainWindow::createTab(int index)
 
     m_tabStates.insert(index, newTab);
     m_tabBar->insertTab(index, tr("Loading..."));
+    m_tabBar->setCurrentIndex(index);
+}
+
+void MainWindow::duplicateTab(int index)
+{
+    if (index < 0 || index >= m_tabStates.size())
+        return;
+
+    TabState *previous = m_tabStates.at(index++);
+    TabState *newTab = new TabState(*previous);
+
+    connect(newTab->searchModel, &SearchModel::queryCompleted, this, &MainWindow::queryCompleted);
+    connect(newTab->tocModel, &SearchModel::queryCompleted, this, &MainWindow::toggleToc);
+
+    m_tabStates.insert(index, newTab);
+    m_tabBar->insertTab(index, newTab->title());
     m_tabBar->setCurrentIndex(index);
 }
 
--- a/src/ui/mainwindow.h
+++ b/src/ui/mainwindow.h
@@ -93,6 +93,7 @@ private slots:
     void openDocset(const QModelIndex &index);
     void queryCompleted();
     void closeTab(int index = -1);
+    void duplicateTab(int index);
 
 private:
     void displayViewActions();


trollixx:
Thanks for the contribution!

I can see a few minor issues, that should be pretty easy to fix. I'll add detailed comments in the code. Also, I am planning to push a few changes I have, so this PR might need a rebase.

src/ui/mainwindow.cpp
```diff
@@ -102,6 +114,10 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     connect(focusSearch, &QShortcut::activated,
             ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
 
+    QShortcut *duplicate = new QShortcut(QKeySequence(QStringLiteral("Ctrl+Alt+T")), this);
+    duplicate->setContext(Qt::ApplicationShortcut);
```
trollixx: No need in `Qt::ApplicationShortcut`.

src/ui/mainwindow.cpp
```diff
@@ -102,6 +114,10 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     connect(focusSearch, &QShortcut::activated,
             ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
 
+    QShortcut *duplicate = new QShortcut(QKeySequence(QStringLiteral("Ctrl+Alt+T")), this);
```
trollixx: `QKeySequence` can be constructed implicitly.

src/ui/mainwindow.cpp
```diff
@@ -455,6 +471,36 @@ void MainWindow::createTab()
     m_tabBar->setCurrentIndex(index);
 }
 
+void MainWindow::duplicateTab()
+{
+    saveTabState();
+    TabState* newTab = new TabState(*m_currentTabState);
+    const QString& parentTitle = m_tabBar->tabText(m_tabBar->currentIndex());
+    const QUrl& parentUrl = m_currentTabState->webPage->mainFrame()->url();
```
trollixx: Cannot be compiled with Qt WebEngine.

src/ui/mainwindow.cpp
```diff
@@ -455,6 +471,36 @@ void MainWindow::createTab()
...
+#else
+    newTab->webPage->setLinkDelegationPolicy(QWebPage::DelegateExternalLinks);
+    newTab->webPage->setNetworkAccessManager(m_application->networkManager());
+    newTab->webPage->mainFrame()->load(parentUrl);
+#endif
+
+    ui->treeView->verticalScrollBar()->setValue(m_currentTabState->searchScrollPosition);
+    ui->tocListView->verticalScrollBar()->setValue(m_currentTabState->tocScrollPosition);
+
+    m_tabStates.append(newTab);
```
trollixx: This breaks if the tab to duplicate is not the last one. Indices of states in `m_tabStates` have to match their tabs.

src/ui/mainwindow.cpp
```diff
@@ -455,6 +471,36 @@ void MainWindow::createTab()
     m_tabBar->setCurrentIndex(index);
 }
 
+void MainWindow::duplicateTab()
```
trollixx: A lot of code is taken from `MainWindow::createTab()`. It would be nice to avoid this code duplication.

innermous: Aint it gonna get messy a bit if we combine two functionalities in one method?

src/ui/mainwindow.cpp
innermous: Can you explain, what you mean? It works correctly as it should, duplicated tabs indices match their position in tab bar...

src/ui/mainwindow.cpp
```diff
@@ -455,6 +458,24 @@ void MainWindow::createTab()
...
+{
+    saveTabState();
+    TabState* previous = currentTabState();
+
+    createTab();
```
trollixx: I don't like this approach, creating something just to override. I am thinking about extending `createTab()` with the necessary functionality...

trollixx:
An important piece appears to be missing - browser history. The new tab should preserve history of the originating tab.

innermous:
Added history preserving. Still have no idea how to implement all in 1 method though.
Only possible outcome that I see - extend createTab with bool duplicate. But that yields lots of if statements as I see it.
If you have any thought - share please.

src/ui/mainwindow.cpp
```diff
@@ -474,6 +501,28 @@ void MainWindow::createTab()
...
+    *newTab = *previous;
+
+#ifdef USE_WEBENGINE
+    newTab->webPage->setUrl(previous->webPage->url());
+#else
+    newTab->webPage->mainFrame()->setUrl(previous->webPage->mainFrame()->url());
+#endif
+
+    QByteArray historyArray;
+    previous->saveHistory(historyArray);
```
trollixx: I think `TabState`'s copy constructor/assignment should take care of history.

trollixx:
@innermous any reason not to rebase this on top of master? I've made some changes to minify the tab related code. `duplicateTab(int index)` where `index` belongs to the tab to duplicate, would work great. Now, with the minimised amount of code there's no need to merge `createTab()` and `duplicateTab()`.

innermous:
@trollixx sorry, i was trying to understand how to make this work after you removed reloadTabState, but I don't know really.

trollixx:
It should just work automatically, after you call `m_tabBar->setCurrentIndex(index)`.

innermous:
@trollixx i don't understand what's the point of index in this case, if we duplicate current tab.

src/ui/forms/mainwindow.ui
```diff
@@ -140,7 +140,8 @@
               <string>←</string>
              </property>
              <property name="icon">
-              <iconset theme="go-previous"/>
+              <iconset theme="go-previous">
+               <normaloff>.</normaloff>.</iconset>
```
trollixx: This and similar changes below are irrelevant.

src/ui/forms/mainwindow.ui
```diff
@@ -311,12 +318,18 @@
...
    <property name="icon">
-    <iconset theme="edit-find"/>
+    <iconset theme="edit-find">
+     <normaloff>.</normaloff>.</iconset>
    </property>
    <property name="text">
     <string>&amp;Find</string>
    </property>
   </action>
+  <action name="actionDuplicateTab">
```
trollixx: I've checked with other browsers, and neither has `Duplicate Tab` in the main menu. I think there's no need to add it here. Its place is tab's context menu, which is yet to be implemented. Just a shortcut should be fine for now.

src/ui/mainwindow.cpp
```diff
@@ -468,6 +514,22 @@ void MainWindow::createTab(int index)
     m_tabBar->setCurrentIndex(index);
 }
 
+void MainWindow::duplicateTab(int index)
```
trollixx: Let's make `index` a required parameter here, and actually duplicate the specific tab. That will allow duplicating tabs from a context menu, once we have it.

src/ui/mainwindow.cpp
```diff
@@ -468,6 +514,22 @@ void MainWindow::createTab(int index)
...
+    if (index == -1)
+        index = m_tabStates.size();
+
+    TabState* previous = currentTabState();
+    TabState* newTab = new TabState(*previous);
+
+    connect(newTab->searchModel, &SearchModel::queryCompleted, this, &MainWindow::queryCompleted);
+    connect(newTab->tocModel, &SearchModel::queryCompleted, this, &MainWindow::toggleToc);
+
+    m_tabStates.insert(index, newTab);
```
trollixx: The new tab should be placed after the one we duplicate.

src/ui/mainwindow.cpp
```diff
@@ -85,6 +85,40 @@ struct TabState
...
+        restoreHistory(historyArray);
+    }
+
+    void restoreHistory(const QByteArray &data) const
+    {
+        QDataStream stream(data);
+        stream >> *webPage->history();
+    }
+
+    void saveHistory(QByteArray &data) const
```
trollixx: Just for the sake of consistency, let's follow Qt's pattern (see `QSplitter::saveState()`, `QByteArray QWidget::saveGeometry()`) here:

```cpp
    QByteArray saveHistory() const
    {
        QByteArray history;
        QDataStream stream(&history, QIODevice::WriteOnly);
        stream << *webPage->history();
        return history;
    }
```

src/ui/mainwindow.cpp
```diff
@@ -85,6 +85,40 @@ struct TabState
...
+        webPage->setLinkDelegationPolicy(QWebPage::DelegateExternalLinks);
+        webPage->setNetworkAccessManager(Core::Application::instance()->networkManager());
+#endif
+
+        QByteArray historyArray;
+        rhs.saveHistory(historyArray);
+        restoreHistory(historyArray);
+    }
+
+    void restoreHistory(const QByteArray &data) const
```
trollixx:
```cpp
void restoreHistory(const QByteArray &history) const
```
Nitpicking :)

trollixx:
Sorry, it took me so long to get back. Things should go better from now on. Thank you for bearing with me! :)

src/ui/mainwindow.cpp
innermous: What's wrong in 
```cpp
 void restoreHistory(const QByteArray &history) const
```
?

src/ui/mainwindow.h
```diff
@@ -93,6 +93,7 @@ private slots:
     void openDocset(const QModelIndex &index);
     void queryCompleted();
     void closeTab(int index = -1);
+    void duplicateTab(int index = -1);
```
trollixx: There should be no default value now.

src/ui/mainwindow.cpp
```diff
@@ -468,6 +515,19 @@ void MainWindow::createTab(int index)
     m_tabBar->setCurrentIndex(index);
 }
 
+void MainWindow::duplicateTab(int index)
+{
+    TabState* previous = m_tabStates.at(index++);
```
trollixx: I'd still check whether the index is valid.

trollixx: And please use `TabState *previous` style.

trollixx:
A few tiny things above, a rebase, and it's good to go!

And please amend the commit message to `ui: Implement tab duplication (fixes #524)`. Thanks!

PR #539, opened 2016-04-20
Implemented opening new tab after current one

innermous:
In addition to previous PR, this feature is good usability enhancement, for me at least. I don't know your personal point of view on it, but maybe you can find it useful.

### DIFF
--- a/src/core/settings.cpp
+++ b/src/core/settings.cpp
@@ -41,6 +41,7 @@ namespace {
 const char GroupBrowser[] = "browser";
 const char GroupDocsets[] = "docsets";
 const char GroupGlobalShortcuts[] = "global_shortcuts";
+const char GroupTabs[] = "tabs";
 const char GroupInternal[] = "internal";
 const char GroupState[] = "state";
 const char GroupProxy[] = "proxy";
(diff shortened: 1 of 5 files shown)


trollixx:
Hey, yeah, this is definitely a required feature. I have several concerns about the implementation.

What I've been keeping in my mind is a more advanced tab management. [Tab Mix Plus](https://addons.mozilla.org/en-US/firefox/addon/tab-mix-plus/) is a good source of inspiration. Adding new setting should be done carefully, because changing anything later will cause some headache with migrations.

I'd like to design the corresponding settings group in full prior to adding individual features. I'll go over this on weekend, and provide more details. Also some changes I am making in `MainWindow` should make everything easier and cleaner.

Thanks for your work!

innermous:
@trollixx okay, looking forward to it!

src/core/settings.cpp
```diff
@@ -41,6 +41,7 @@ namespace {
 const char GroupBrowser[] = "browser";
 const char GroupDocsets[] = "docsets";
 const char GroupGlobalShortcuts[] = "global_shortcuts";
+const char GroupTabsBehavior[] = "tabs_behavior";
```
trollixx: Please change this into `tabs`.

trollixx:
Please rebase this on top of master, and change the commit message to `core, ui: Add option for new tabs to be open next to current one`. Thanks.

trollixx:
Merged, thanks!

Issue #546, opened 2016-05-05
Run once per X session


chris-morgan:
I routinely keep two X sessions active: one for my own personal stuff, one for my work stuff, which I keep mostly disjoint.

Most apps that try to keep only one instance active make a hash of this situation. Zeal is no exception. (Firefox is a great example of an app that handles this scenario well. I can run one instance with one profile and another with another.)

Zeal’s checks for an existing instance should, in X, only search for instances running on the current display.

trollixx:
That is a pretty extreme use case of yours. Allowing one application instance per X session might be possible, but then support for multiple profiles has to be implemented first.

Issue #78, opened 2013-12-21
Save Workspace


mazen-mardini:
It would be nice if the workspace on Zeal could be saved. All my expanded docsets and browsing history should be saved when quitting Zeal.

geonicz:
Really bored of reopening tabs after reboot... Arch linux, quick kernel changes, notebook - reboot is a must. All tabs are lost. Disturbing.

cocoonkid:
+1

PR #548, opened 2016-05-16
Implement restoring last session

innermous:
I need one thing from #538 to be rebased to make tab titles work properly. 
UPD. Forgot about commit messages style, will fix that with next patch set.

### DIFF
--- a/src/core/settings.cpp
+++ b/src/core/settings.cpp
@@ -77,6 +77,7 @@ void Settings::load()
     // TODO: Put everything in groups
     startMinimized = m_settings->value(QStringLiteral("start_minimized"), false).toBool();
     checkForUpdate = m_settings->value(QStringLiteral("check_for_update"), true).toBool();
+    restoreLastSession = m_settings->value(QStringLiteral("restore_last_state"), false).toBool();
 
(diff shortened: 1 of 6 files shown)


innermous:
By the way. What if we save session every time application is closed?
This can give user two options 
* auto restore it on startup if user checked "Restore last session" in settings
* restore it explicitly when user explicitly wants it with Ctrl+Shift+T like browsers do

trollixx:
This one is tricky. I've been thinking about allowing people to have several sessions similarly to Qt Creator's sessions or or even Firefox's profiles (see #546 too). I guess we can start with this, and then add more features.

What's actually bothering me, is adding more stuff to `MainWindow`. I've been working on moving things out, so all UI code could be eventually encapsulated in a plugin. How about coming up with an independent class (e.g. `SessionManager`), and letting it do all dirty work?

trollixx:
Adding a reference to #78, which is the issue this patch attempts to resolve.

innermous:
@trollixx I've never seen that issue, just noticed a tiny ToDo in the code. Thanks.

trollixx:
Outdated, and requires a rewrite.

PR #564, opened 2016-07-18
Defer opening the first query result

mrhota:
Add a timer that defers the browser opening.
This prevents from loading many pages as a user types the query.

### DIFF
--- a/src/ui/mainwindow.cpp
+++ b/src/ui/mainwindow.cpp
@@ -42,6 +42,7 @@
 #include <QShortcut>
 #include <QSystemTrayIcon>
 #include <QTabBar>
+#include <QTimer>
 
 #ifdef USE_WEBENGINE
 #include <QWebEngineHistory>
@@ -174,6 +175,7 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     m_application(app),
     m_settings(app->settings()),
     m_zealListModel(new ListModel(app->docsetRegistry(), this)),
+    m_deferOpenUrl(new QTimer()),
     m_globalShortcut(new QxtGlobalShortcut(m_settings->showShortcut, this))
 {
     ui->setupUi(this);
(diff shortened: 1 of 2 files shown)


mrhota:
@trollixx r?

trollixx:
Not until I get home. I'll take a look tonight.

src/ui/mainwindow.cpp
```diff
@@ -174,6 +175,7 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     m_application(app),
     m_settings(app->settings()),
     m_zealListModel(new ListModel(app->docsetRegistry(), this)),
+    m_deferOpenUrl(new QTimer()),
```
trollixx: `new QTimer(this)`

src/ui/mainwindow.h
```diff
@@ -122,6 +123,8 @@ private slots:
     QMenu *m_backMenu = nullptr;
     QMenu *m_forwardMenu = nullptr;
 
+    QTimer *m_deferOpenUrl;
```
trollixx: It would be nice to have a more sensible name here, at least like `m_openDocsetTimer` or so. Also please add `= nullptr`.

src/ui/mainwindow.cpp
```diff
@@ -466,6 +468,21 @@ void MainWindow::search(const SearchQuery &query)
     ui->treeView->activated(ui->treeView->currentIndex());
 }
 
+void MainWindow::deferOpenDocset(const QModelIndex &index)
+{
+    /// PERF: Loading a web page makes the input sluggish.
```
trollixx: Please no Doxygen `///` comment style.

trollixx:
While testing this patch, I came up with a bit more efficient implementation, which I have committed.

Thank you for pushing me to get this improvement in. :)

PR #579, opened 2016-08-09
Tab alignment fix

spegoraro:
The bottom of the ToolBarFrame was out of alignment with the bottom of the search text field and it was driving me nuts!

This PR resolves it by using a vertical spacer in the ToolBarFrame which expands the widget to a maximum height matching that of the search bar's container. I also modified the color of the bottom line to the palette's mid color to suit nicer with most current themes.

Here's a comparison:

Before
zeal_before

After
zeal_after

Thanks for the awesome work on Zeal!

### DIFF
--- a/src/ui/forms/mainwindow.ui
+++ b/src/ui/forms/mainwindow.ui
@@ -29,11 +29,23 @@
     </property>
     <item>
      <widget class="QSplitter" name="splitter">
+      <property name="sizePolicy">
+       <sizepolicy hsizetype="Expanding" vsizetype="Ignored">
+        <horstretch>0</horstretch>
+        <verstretch>0</verstretch>
+       </sizepolicy>
+      </property>
       <property name="lineWidth">
        <number>0</number>
       </property>
       <property name="orientation">
        <enum>Qt::Horizontal</enum>
+      </property>
+      <property name="opaqueResize">
+       <bool>true</bool>
+      </property>
+      <property name="handleWidth">
+       <number>1</number>
       </property>
       <property name="childrenCollapsible">
        <bool>false</bool>
@@ -84,6 +96,12 @@
          <widget class="QSplitter" name="tocSplitter">
           <property name="orientation">
            <enum>Qt::Vertical</enum>
+          </property>
+          <property name="opaqueResize">
+           <bool>true</bool>
+          </property>
+          <property name="handleWidth">
+           <number>1</number>
           </property>
           <widget class="QTreeView" name="treeView">
            <property name="frameShape">
--- a/src/ui/widgets/toolbarframe.cpp
+++ b/src/ui/widgets/toolbarframe.cpp
@@ -13,7 +13,7 @@ void ToolBarFrame::paintEvent(QPaintEvent *event)
     // Draw a line at the bottom.
     QPainter painter(this);
     painter.save();
-    painter.setPen(palette().dark().color());
+    painter.setPen(palette().mid().color());
     painter.drawLine(0, height() - 1, width(), height() - 1);
     painter.restore();
 }


bendem:
This breaks the forward/backward buttons here.

Before


After

groksteve:
Not getting that here.
As of the latest master the buttons don't have 'Back' and 'Forward' text any more like in your before screenshow, they have arrow icons. Please try a fresh clone of zealdocs/zeal and let me know how it goes.

bendem:
This is a clean HEAD (ebd3ab8).

src/ui/forms/mainwindow.ui
```diff
@@ -29,12 +29,24 @@
...
       <property name="lineWidth">
        <number>0</number>
       </property>
       <property name="orientation">
        <enum>Qt::Horizontal</enum>
       </property>
+      <property name="opaqueResize">
+       <bool>true</bool>
+      </property>
+      <property name="handleWidth">
```
trollixx: Please do not modify `handleWidth`, it should depend on system's style.

src/ui/forms/mainwindow.ui
```diff
@@ -82,6 +97,12 @@
           <property name="orientation">
            <enum>Qt::Vertical</enum>
           </property>
+          <property name="opaqueResize">
+           <bool>true</bool>
+          </property>
+          <property name="handleWidth">
```
trollixx: Same here.

src/ui/forms/mainwindow.ui
```diff
@@ -29,9 +29,24 @@
...
+        <verstretch>0</verstretch>
+       </sizepolicy>
+      </property>
+      <property name="lineWidth">
+       <number>0</number>
+      </property>
       <property name="orientation">
        <enum>Qt::Horizontal</enum>
       </property>
+      <property name="opaqueResize">
```
trollixx: `opaqueResize` is style dependent by default, and should stay this way.

src/ui/forms/mainwindow.ui
```diff
@@ -29,9 +29,24 @@
     </property>
     <item>
      <widget class="QSplitter" name="splitter">
+      <property name="sizePolicy">
+       <sizepolicy hsizetype="Expanding" vsizetype="Ignored">
+        <horstretch>0</horstretch>
+        <verstretch>0</verstretch>
+       </sizepolicy>
+      </property>
+      <property name="lineWidth">
```
trollixx: Where does this come from? `QSplitter` doesn't have `lineWidth` property.

trollixx: Found it, `QFrame`, looks like it's also supposed to vary based on style.

src/ui/mainwindow.cpp
```diff
@@ -636,7 +636,7 @@ void MainWindow::setupTabBar()
     m_tabBar->setDrawBase(false);
     m_tabBar->setDocumentMode(true);
     m_tabBar->setElideMode(Qt::ElideRight);
-    m_tabBar->setStyleSheet(QStringLiteral("QTabBar::tab { width: 150px; }"));
+    m_tabBar->setStyleSheet(QStringLiteral("QTabBar::tab { width: 150px; height: 39px; }"));
```
trollixx: Setting tab height is a bad idea, as different themes have different tab heights.

trollixx:
Thanks for the contribution. It's indeed an issue with some user themes, but my goal is to rely on system's style as much as possible. I've pointed out a few things that should not be changed, but there might be more. The better approach would be to use maximum of heights of the left and the right bars.

groksteve:
Ok thanks, I'll clean those up. Btw I'm fairly new to Qt, could you point me to where I can find which properties are best left to the theme?

trollixx:
> could you point me to where I can find which properties are best left to the theme?

It's all in the [docs](http://doc.qt.io/qt-5/). For example, [QSplitter::handleWidth](http://doc.qt.io/qt-5/qsplitter.html#handleWidth-prop) says:
> By default, this property contains a value that depends on the user's platform and style preferences.

groksteve:
Thanks! My evening lazy skimming over the docs missed that!

trollixx:
Should be fixed by 2fd6ea124e88b6629540c06f6ed3dc51cd501308.

PR #623, opened 2016-10-09
ui: Fix docset removing on Windows (fixes #621)

### LINKED ISSUE #621: Zeal 0.3.0 remove docset not actually working
In 0.3.0 when remove a docset int the docset manager UI, it gets removed from the list and the main UI. However  files on the disk remain untouched and after restart Zeal the deleted docset just reappear in the list.

My system is Win10, tried portable and non-portable version.

jkozera:
Docsets need to be removed from registry first, to avoid rename() failing when files are in use.

### DIFF
--- a/src/libs/ui/docsetsdialog.cpp
+++ b/src/libs/ui/docsetsdialog.cpp
@@ -693,11 +693,19 @@ void DocsetsDialog::removeDocset(const QString &name)
     const QString tmpPath = docsetPath + QLatin1String(".deleteme.")
             + QString::number(QDateTime::currentMSecsSinceEpoch());
 
-    // Rename first to allow simultaneous installation.
-    // TODO: Check for error
-    QDir().rename(docsetPath, tmpPath);
-
+    // Remove from registry first to avoid renaming files in use on Windows.
     m_docsetRegistry->remove(name);
+
+    // Rename first to allow simultaneous installation.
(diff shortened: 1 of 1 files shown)


src/libs/ui/docsetsdialog.cpp
```diff
@@ -693,12 +693,21 @@ void DocsetsDialog::removeDocset(const QString &name)
     const QString tmpPath = docsetPath + QLatin1String(".deleteme.")
             + QString::number(QDateTime::currentMSecsSinceEpoch());
 
-    // Rename first to allow simultaneous installation.
-    // TODO: Check for error
-    QDir().rename(docsetPath, tmpPath);
-
+    // Remove from registry first to avoid renaming files in use on Windows
```
trollixx: Missing period! :)

src/libs/ui/docsetsdialog.cpp
```diff
@@ -693,12 +693,21 @@ void DocsetsDialog::removeDocset(const QString &name)
...
     m_docsetRegistry->remove(name);
 
+    // Rename first to allow simultaneous installation.
+    if (!QDir().rename(docsetPath, tmpPath)) {
+        const QString error = tr(
```
trollixx:
```cpp
const QString error = tr("Cannot delete docset <b>%1</b>! Please try closing other "
                         "applications first, as they may be accessing the docset "
                         "files.").arg(title);
```

PR #660, opened 2016-11-23
Allow download in background

haobug:
Allow user to search while downloading, deleting, or do other operation in background.

### DIFF
--- a/src/libs/core/application.cpp
+++ b/src/libs/core/application.cpp
@@ -29,6 +29,7 @@
 #include <registry/docsetregistry.h>
 #include <registry/searchquery.h>
 #include <ui/mainwindow.h>
+#include <ui/docsetsdialog.h>
 #include <util/version.h>
 
 #include <QCoreApplication>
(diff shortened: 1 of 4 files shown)


trollixx:
This is basically a reversal of a very old change. Unfortunately, it's known to cause crashes due to race conditions, e.g. when you're performing a search within a docset that's being deleted. So, I cannot merge this.

The right approach here is to move all download/extract/delete logic to `Core`, or even a new module. Also docsets that are about to be deleted, or updated should be immediately removed/hidden from the UI.

Let me know in case you'd like to work on a better docset management system.

haobug:
> The right approach here is to move all download/extract/delete logic to Core

yes indeed. the curent implemntation handling the download and progress status in ui object of docsetDialog, i did that in this way for minimum code changes. 

>or even a new module. Also docsets that are about to be deleted, or updated should be immediately removed/hidden from the UI.
>Let me know in case you'd like to work on a better docset management system.

i don't know must about the underhood of zeal docset management.
i'd like to do that in small pieces. i will first try to look into some docset related defect.

PR #661, opened 2016-11-23
add shortcut guide user to tools menu

haobug:
add a shortcut on docsets tab of the settings dialog
to guide user to the new entry of docsets dialog.

### DIFF
--- a/src/libs/core/application.cpp
+++ b/src/libs/core/application.cpp
@@ -219,6 +219,11 @@ QString Application::userAgent()
     return QStringLiteral("Zeal/%1").arg(QCoreApplication::applicationVersion());
 }
 
+void Application::showToolsMenu()
+{
+    this->m_mainWindow->showToolsMenu();
+}
+
 QString Application::userAgentJson() const
 {
     // TODO: [Qt 5.4] Remove else branch
--- a/src/libs/core/application.h
+++ b/src/libs/core/application.h
@@ -56,6 +56,7 @@ public:
     Settings *settings() const;
 
     Registry::DocsetRegistry *docsetRegistry();
+    void showToolsMenu();
 
 public slots:
     void executeQuery(const Registry::SearchQuery &query, bool preventActivation);
--- a/src/libs/ui/forms/settingsdialog.ui
+++ b/src/libs/ui/forms/settingsdialog.ui
@@ -20,7 +20,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>4</number>
      </property>
      <widget class="QWidget" name="generalTab">
       <attribute name="title">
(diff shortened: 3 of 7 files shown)


src/libs/core/application.h
```diff
@@ -56,6 +56,7 @@ class Application : public QObject
     Settings *settings() const;
 
     Registry::DocsetRegistry *docsetRegistry();
+    void showToolsMenu();
```
trollixx: The `Core` module should not depend on `Ui`. Currently there's a dependency on `Ui::MainWindow`, but that will go away soon. So, `Application` definitely cannot be responsible for showing a menu.

src/libs/ui/forms/settingsdialog.ui
```diff
@@ -20,7 +20,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>4</number>
```
trollixx: You have left the last tab selected.

src/libs/ui/forms/settingsdialog.ui
```diff
@@ -402,6 +402,16 @@
          </property>
         </widget>
        </item>
+       <item>
+        <widget class="QPushButton" name="showToolsMenu">
+         <property name="locale">
```
trollixx: No need to select locale here.

src/libs/ui/mainwindow.cpp
```diff
@@ -907,6 +907,12 @@ void MainWindow::applySettings()
     else
         removeTrayIcon();
 }
+void MainWindow::showToolsMenu()
+{
+    QMenuBar *menubar = this->menuBar();
+    QMenu *menu_Tools = menubar->findChild<QMenu *>("menu_Tools");
+    menubar->setActiveAction(menu_Tools->menuAction());
```
trollixx: `ui->menu_Tools` instead of the magic above. :)

src/libs/ui/settingsdialog.cpp
```diff
@@ -70,6 +72,13 @@ SettingsDialog::~SettingsDialog()
     delete ui;
 }
 
+void SettingsDialog::on_showToolsMenu_clicked()
+{
+    this->saveSettings();
+    this->reject();
+    m_application->showToolsMenu();
```
trollixx: This breaks encapsulation, you need to fire a signal instead of calling a method directly. And then simply connect to that signal in `MainWindow`.

src/libs/core/application.cpp
```diff
@@ -219,6 +219,11 @@ QString Application::userAgent()
     return QStringLiteral("Zeal/%1").arg(QCoreApplication::applicationVersion());
 }
 
+void Application::showToolsMenu()
+{
+    this->m_mainWindow->showToolsMenu();
```
trollixx: No need in `this->` in C++.

trollixx:
Hi, thanks for the contribution. I am not quite sure about this, because I am going to delete the whole tab soon, so there's no much point to make changes now...

haobug:
learned a lot from you comment. i will update this pr later.
you can leave this pr open untill you remove the docsets tab.

PR #669, opened 2016-12-15
add macOS build support

Akagi201:


trollixx:
There's [a wiki page](https://github.com/zealdocs/zeal/wiki/Build-Instructions-for-OS-X) that describes how to build Zeal on macOS.

I don't like the idea of hardcoding Homebrew specifc paths. That might also imply that we support macOS, which is untrue. It just happens to build there, but that may be not true tomorrow.

PR #670, opened 2016-12-15
ProgressItemDelegate fixes

htower:
The patch fixes delegate's progress bar painting. Old version don't show progress bar if text width (depends on font size) is close to allowed maximum (`progressBarWidth` constant value is too small) and also paint progress bar with direct using of `QProgressBar` class - this leads to painting default widget background color (instead needed transparent).

New version fixes these problems by using `QStyleOptionProgressBar` class and increasing `progressBarWidth` constant value.

### DIFF
--- a/src/libs/ui/docsetsdialog.cpp
+++ b/src/libs/ui/docsetsdialog.cpp
@@ -127,7 +127,7 @@ DocsetsDialog::DocsetsDialog(Core::Application *app, QWidget *parent) :
         ui->availableDocsetList->selectionModel()->select(index, QItemSelectionModel::Deselect);
 
         QAbstractItemModel *model = ui->availableDocsetList->model();
-        model->setData(index, tr("Downloading: %p%"), ProgressItemDelegate::FormatRole);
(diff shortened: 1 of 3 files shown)


htower:
Maybe we should remove percentage info from delegate and show only text like "Downloading"/"Installing"? Then we can fix progress bar size and place text (right-aligned) at left, something like (`_` is for align):
```
docset1__Downloading [====     ]
docset2___Installing [=======  ]
```

trollixx:
I guess it's a Windows specific problem (or a theme specific), because with my current style text is placed within a progress bar. Not saying that we shouldn't fix the problem, just need to test how it looks like with different styles.

htower:
> I guess it's a Windows specific problem (or a theme specific)

Sample for linux with standard Breeze theme (with `progressBarWidth = 300`):
sample

`QStyleOptionFOO` classes is the recommended way to draw controls inside delegates, all such samples uses them.

trollixx:
Docset Dialog will undergo a major revamp, no for this fix. Thanks anyway.

PR #672, opened 2016-12-21
registry: Cleanup fuzzy search code, and optimize a bit

trollixx:
An attempt to make it a bit cleaner, and, hopefully, more efficient. I have a few more optimizations in mind, but let's start with these.

Please give this a look, and maybe some testing/benchmarking.


trollixx:
All right, nothing complex here, let's test in production, as usual :)

PR #718, opened 2017-04-14
registry: Add more exhaustive search for fuzzy matching

jkozera:
Fixes 2 search cases not matching with fuzzy search:

1. "**ad**darg" not matching  "**a**rgparse.ArgumentParser.a**d**d_argument" in the Python docset,
2. "**te**strle" not matching "AWS.Elas**t**icTranscod**e**r.testRole" in the AWS JS docset.

Both were due to the 2nd char of needle being more than 8 chars from 1st match of 1st needle char in haystack (highlighted with **bold** above). Adding a simple recursive call to retry matching from 2nd match of 1st char fixes both cases.

It is slightly slower now - my simple performance test case of [all 2M symbols from all docsets](https://github.com/jkozera/zeal/blob/c30da109e8624b961e7b94f3f59f783b188c0d91/tests/py_unit/testscoring_build.py#L35) indicates about 20% slow down, from 1.2 seconds to 1.45 seconds on my box, on the "aaaaaaaaaa" query (which should be rather heavy, since it should involve scanning all the symbols multiple times due to lack of matches).

This investigation was inspired by the [recursive matching from lib_fts](https://github.com/forrestthewoods/lib_fts/blob/d541eb49b289a03320fb62150b8fd01364e22403/code/fts_fuzzy_match.h#L121), but for now I've decided to go with a simplified implementation, since in general it would be even slower. Preliminary testing with recursive call on each character match shows the slowdown would be like 40% instead of 20%, so maybe still feasible, but I'd like to try the faster implementation first. Especially since I'm also not sure if recursing on each match would actually improve the results - so if someone can find a reasonable test case which still doesn't work here, I'd reconsider. :)

### DIFF
--- a/src/libs/registry/docset.cpp
+++ b/src/libs/registry/docset.cpp
@@ -628,6 +628,27 @@ QString Docset::parseSymbolType(const QString &str)
     return aliases.value(str, str);
 }
 
+/**
+ * \brief Returns score based on a substring position in a string.
+ * \param str Original string.
+ * \param index Index of the substring within \a str.
+ * \param length Substring length.
+ * \return Score value between 1 and 100.
+ */
+static int scoreFuzzy(const char *str, int index, int length)
+{
+    if (index == 0 || str[index - 1] == '.') {
+        // score between 66..99, if the match follows a dot, or starts the string
+        return qMax(66, 100 - length);
+    } else if (str[index + length] == 0) {
+        // score between 33..66, if the match is at the end of the string
+        return qMax(33, 67 - length);
+    } else {
+        // score between 1..33 otherwise (match in the middle of the string)
+        return qMax(1, 34 - length);
+    }
+}
+
 // Based on https://github.com/bevacqua/fuzzysearch
 static void matchFuzzy(const char *needle, int needleLength,
                        const char *haystack, int haystackLength,
@@ -639,6 +660,9 @@ static void matchFuzzy(const char *needle, int needleLength,
     *start = -1;
 
     int groupCount = 0;
+    int bestRecursiveScore = -1;
+    int bestRecursiveStart = -1;
+    int bestRecursiveLength = -1;
 
     for (int i = 0, j = 0; i < needleLength; ++i) {
         bool found = false;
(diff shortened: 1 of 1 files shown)


src/libs/registry/docset.cpp
```diff
@@ -639,6 +660,9 @@ static void matchFuzzy(const char *needle, int needleLength,
     *start = -1;
 
     int groupCount = 0;
+    int bestRecScore = -1;
```
trollixx: Please do not use ambiguous shortenings. Recursive? Record?

src/libs/registry/docset.cpp
```diff
@@ -647,9 +671,24 @@ static void matchFuzzy(const char *needle, int needleLength,
 
         while (j < haystackLength) {
             if (needle[i] == haystack[j++]) {
-                if (*start == -1)
+                if (*start == -1) {
                     *start = j;  // first matched char
 
+                    // try starting the search later in case the first character occurs again later
+                    int recStart, recLength, recScore;
```
trollixx: Same as above.

src/libs/registry/docset.cpp
```diff
@@ -647,9 +671,24 @@ static void matchFuzzy(const char *needle, int needleLength,
...
+                    matchFuzzy(
```
trollixx: Don't break the line immediately after parenthesis, there's quite a lot of space on the same line (up to 100 characters).

src/libs/registry/docset.cpp
```diff
@@ -674,10 +713,23 @@ static void matchFuzzy(const char *needle, int needleLength,
...
+                *start = bestRecStart;
+                *length = bestRecLength;
+            } else {
+                *start = -1;
+            }
             return;
         }
     }
+
+    int score = scoreFuzzy(haystack, *start, *length);
```
trollixx: Why re-score again? This case should've been considered by now. Or not?

jkozera: Above only the cases of some `recStart > *start` were considered - here we compare all the later-starting scores against the 'earliest' score. For example, when searching for in "foobaz.foobar" could yield two different scores - matching "[**foo**]baz.foo[**bar**]" or "foobaz.[**foobar**]" - in the latter case of exact match the score would be larger.

I hope to address the other comments later today.

Issue #401, opened 2015-09-10
Make Control+L focus the search bar (like browsers)


sunaku:
Please make <kbd>Control</kbd>+<kbd>L</kbd> focus the search bar, like how Web browsers focus their URL address bar.

See http://www.howtogeek.com/114518/47-keyboard-shortcuts-that-work-in-all-web-browsers/

trollixx:
Thanks, I'll add it shortly. Currently `Escape` does that, I cannot say I am happy with the choice.

Making Zeal behave more like traditional web browsers is only of my goals. Also, configurable shortcuts could be a good option as well.

Related issues: #10 and #62.

trollixx:
I've just discovered that we also have <kbd>?</kbd> shortcut to select the current query without prefix. Not sure if anyone is aware about this. :)

I am implementing <kbd>Ctrl+L</kbd> right now. It will select the current search term without prefix (if any), and then select everything on the second press.

trollixx:
LOL, further investigation has shown that there's also <kbd>Ctrl+K</kbd> for the same purpose.

sunaku:
From *NIX perspective, <kbd>?</kbd> is also strange.  I would expect <kbd>/</kbd> to search instead (it works in Firefox). :sweat_smile:

trollixx:
After some thinking, I have become less confident in this particular case. Firefox and Chromium support <kbd>F6</kbd>, <kbd>Alt+D</kbd>, <kbd>Ctrl+L</kbd> to focus address bar, and <kbd>Ctrl+K</kbd> allows to select search bar (In Chromium it switches address bar to search-only mode).

In Zeal the search bar is really just a search bar, not an address bar. It does not update when one's navigates through pages. To be consistent with browsers keeping <kbd>Ctrl+K</kbd> as the only shortcut is the right way.

I have some plans to add a breadcrumbs panel like the one Dash has. It might be that <kbd>Ctrl+L</kbd> et al. would work for focusing such panel better.

trollixx:
There's also <kbd>Ctrl+E</kbd> for activating the search mode in Chromium and Opera.

So many different options :)

sunaku:
True. :scream_cat: I think that documenting `?` and `/` keys might suffice: I use them regularly now.

Konfekt:
[...]

PR #749, opened 2017-07-14
Automatically load selected docset on keyboard selection without enter or click

stecman:
This fixes one of my main peeves working between Dash and Zeal. Dash automatically loads the selected docset in the tree when changing the selection with the keyboard. Up until now, Zeal has required enter to be pressed to load a docset selected with the keyboard, which additionally moved focus away from the tree.

That is, in Dash the workflow is:

- search something
- use up/down arrows to move to relevant looking item
- [docset loads after a small delay]
- use up/down arrows to move to a different item
- ...

While in Zeal this has been:

- search something
- use up/down arrows to move to an item
- press enter to load selected item
- [focus moves to web view, keyboard can no longer change selection]
- use mouse to pick a different item or to focus tree again
- ...

This commit implements behaviour similar to Dash:

zeal-update

Notable changes:

- Focus is no longer lost (shifted to the web view) when a docset is loaded unless it is actively selected with enter or a click. This is mostly to maintain the existing behaviour for these actions. Personally I feel clicking an item should leave focus in the tree for further keyboard selection.

- The timeout for searching now serves as a general purpose delayed docset load timeout. This limits the typing and selection delays to the same value, but the 400ms currently used seems to work well in both cases.

- An "active" load (click/enter) overrides any delayed load (keyboard selection)

Other notes:

- Dash appears to throttle loading rather than use a fixed delay. Throttling results in a slightly nicer experience as changing the selection slowly loads the new selection without delay.

### DIFF
--- a/src/libs/ui/mainwindow.cpp
+++ b/src/libs/ui/mainwindow.cpp
@@ -328,9 +328,12 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
 
     createTab();
 
-    connect(ui->treeView, &QTreeView::clicked, this, &MainWindow::openDocset);
+    // Focus web view on any active interaction with tree (enter, clicking)
+    connect(ui->treeView, &QTreeView::clicked, this, &MainWindow::openAndFocusDocset);
+    connect(ui->treeView, &QTreeView::activated, this, &MainWindow::openAndFocusDocset);
+
+    // Handle selections in the table of contents list
     connect(ui->tocListView, &QListView::clicked, this, &MainWindow::openDocset);
-    connect(ui->treeView, &QTreeView::activated, this, &MainWindow::openDocset);
     connect(ui->tocListView, &QListView::activated, this, &MainWindow::openDocset);
 
     connect(m_application->docsetRegistry(), &Registry::DocsetRegistry::searchCompleted,
@@ -386,9 +389,6 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
             return;
 
         openDocset(index);
-
-        // Get focus back.
-        ui->lineEdit->setFocus(Qt::MouseFocusReason);
     });
 
     ui->actionNewTab->setShortcut(QKeySequence::AddTab);
@@ -463,13 +463,31 @@ void MainWindow::search(const Registry::SearchQuery &query)
     emit ui->treeView->activated(ui->treeView->currentIndex());
 }
 
+// Open the docset at index immediately
 void MainWindow::openDocset(const QModelIndex &index)
 {
+    // Cancel any other pending docset open
+    m_openDocsetTimer->stop();
+
     const QVariant url = index.data(Registry::ItemDataRole::UrlRole);
     if (url.isNull())
         return;
 
     currentTab()->load(url.toUrl());
+}
+
+// Open the docset at index after a delay (for typing, keyboard selection, etc)
+void MainWindow::openDocsetDelayed(const QModelIndex &index)
+{
+    m_openDocsetTimer->stop();
+    m_openDocsetTimer->setProperty("index", index);
+    m_openDocsetTimer->start();
+}
+
+// Open the docset at index and move focus to the web view
+void MainWindow::openAndFocusDocset(const QModelIndex &index)
+{
+    openDocset(index);
     currentTab()->focus();
 }
 
@@ -493,8 +511,7 @@ void MainWindow::queryCompleted()
 
     ui->treeView->setCurrentIndex(currentTabState()->searchModel->index(0, 0, QModelIndex()));
 
-    m_openDocsetTimer->setProperty("index", ui->treeView->currentIndex());
-    m_openDocsetTimer->start();
+    openDocsetDelayed(ui->treeView->currentIndex());
 }
 
 void MainWindow::closeTab(int index)
@@ -580,8 +597,20 @@ void MainWindow::syncTreeView()
 
     // TODO: Remove once QTBUG-49966 is addressed.
     QItemSelectionModel *newSelectionModel = ui->treeView->selectionModel();
-    if (oldSelectionModel && newSelectionModel != oldSelectionModel) {
-        oldSelectionModel->deleteLater();
+    if (newSelectionModel != oldSelectionModel) {
+        if (oldSelectionModel) {
+            oldSelectionModel->deleteLater();
+        }
+
+        // Connect to new selection model
+        connect(ui->treeView->selectionModel(), &QItemSelectionModel::selectionChanged,
+            this, [this](const QItemSelection &selected) {
+                if (selected.isEmpty()) {
+                    return;
+                }
+
+                openDocsetDelayed(selected.indexes().first());
+            });
     }
 
     ui->treeView->reset();
--- a/src/libs/ui/mainwindow.h
+++ b/src/libs/ui/mainwindow.h
@@ -28,6 +28,7 @@
 
 class QxtGlobalShortcut;
 
+class QItemSelection;
 class QModelIndex;
 class QSystemTrayIcon;
 class QTabBar;
@@ -79,6 +80,8 @@ protected:
 private slots:
     void applySettings();
     void openDocset(const QModelIndex &index);
+    void openAndFocusDocset(const QModelIndex &index);
+    void openDocsetDelayed(const QModelIndex &index);
     void queryCompleted();
     void closeTab(int index = -1);
     void moveTab(int from, int to);


trollixx:
Hi, thanks for contributing!

We are currently preparing the next release, so we can merge this once 0.4.0 is out.

Personally I like to move focus within the tree view without changing contents of the web view. How about making this behavior configurable?

The build errors require fixing, and I'll go over the code shortly.

stecman:
CMake build fixed :+1: I was using `qmake` to build and test this under Linux - didn't realise CMake was in there too with a manually maintained list of source paths.

Would you consider having this behaviour as an option, but with the Dash-like behaviour the default? Not sure how many people have used Dash before, but coming from Dash the Zeal behaviour around this felt very clunky (particuarly the tree losing focus).

Also out of curiosity, what's your process for selecting different items in the tree if this doesn't drive you nuts?

trollixx:
Thanks for fixing CMake build. I am going to drop qmake support of the release.

I don't mind changing the default behavior, and most likely other users wouldn't start complaining about that.

I am not a super active Zeal user myself, and usually do several searches during my work day, so I just move focus with mouse. But I agree, focus moving to the web view after each activation can cause frustration. Considering we don't have an easy way to focus different areas. For that we need to implement behavior similar to <kbd>F6</kbd> in major browsers (somewhat related to #401).

stecman:
> How about making this behavior configurable?

@trollixx I've been mulling over this for a while (and using a build with branch this daily) and having this as a preference still seems unnecessary to me. I generally avoid adding preferences though, so this might just be my own bias (and workflow).

What do you reckon about having a modifier like <kbd>Alt</kbd> <kbd>↑</kbd>/<kbd>↓</kbd> to change selection without loading instead? (On closer inspection, Dash has this behaviour too).

trollixx:
I will try out your change myself, but at the moment it feels like a configuration switch would be convenient. Also having a way to reverse the configured behavior via some key combination seems to be not a bad idea.

src/libs/ui/widgets/treeview.cpp
```diff
@@ -0,0 +1,44 @@
...
+using namespace Zeal::WidgetUi;
+
+TreeView::TreeView(QWidget *parent) :
+    QTreeView(parent)
+{
+}
+
+void TreeView::selectionChanged(const QItemSelection &selected, const QItemSelection &deselected)
+{
+    QTreeView::selectionChanged(selected, deselected);
```
trollixx: There's no need in this `TreeView` class. Selection updates can be obtained via [`QAbstractitemView::selectionModel`](http://doc.qt.io/qt-5/qabstractitemview.html#selectionModel).

stecman: See the comment below that - it's the first approach I tried, but it doesn't work reliably:

> `QTreeView` doesn't normally expose selection changes as a public signal. `QTreeView's`
> selection model is accessible publicly, however the instance this refers to can change
> internally, so it's not reliable to connect these `QItemSelectionModel` signals externally.

The selection model gets switched out regularly in the current implementation, so you end up with a pointer to an old selection model after the first search and the connection no longer reports selection changes.

trollixx:
I am still interested in this, it's just that I have some doubts about the implementation.  Please bear with me, while I find some spare time to play with this.

stecman:
@trollixx I've updated this to remove the `TreeView` subclass. Found the selection model was being replaced every time `TreeView::setModel` was called, so I've moved the `selectionChanged` connect to `MainWindow::syncTreeView`.

Also rebased against master :+1:

trollixx:
While reading about selection models, I realized that the ones replaced after `setModel()` are kept until the application exits. There's an unresolved [QTBUG-49966](https://bugreports.qt.io/browse/QTBUG-49966) report about this situation.

I've fixed the problem in 14330638bb32ea8ea7529466cff44b868c058619, so this PR needs another update, sorry :)

stecman:
Rebased, thanks. I noticed the selection model didn't seem to be cleaned up when looking at the Qt source for this, but didn't look into it further :fire:

src/libs/ui/mainwindow.cpp
```diff
@@ -568,7 +585,12 @@ void MainWindow::syncTreeView()
         oldSelectionModel->deleteLater();
     }
 
+    // Connect to new selection model
+    connect(ui->treeView->selectionModel(), &QItemSelectionModel::selectionChanged,
```
trollixx: Connection should be done only for the new selection model. Plus there's really no need in the `docsetSelectionChanged` slot. A lambda will do just fine.
I suggest rewriting this the following way:

```cpp
    // TODO: Remove once QTBUG-49966 is addressed.
    QItemSelectionModel *newSelectionModel = ui->treeView->selectionModel();
    if (newSelectionModel != oldSelectionModel) {
        if (oldSelectionModel) {
            oldSelectionModel->deleteLater();
        }

        connect(newSelectionModel, &QItemSelectionModel::selectionChanged,
                this, [this](const QItemSelection &selected) {
            if (selected.isEmpty()) {
                return;
            }

            m_openDocsetTimer->start();
            m_openDocsetTimer->setProperty("index", selected.indexes().first());
        });
    }
```

src/libs/ui/mainwindow.cpp
```diff
@@ -568,7 +585,12 @@ void MainWindow::syncTreeView()
...
+        this, &MainWindow::docsetSelectionChanged);
+
     ui->treeView->reset();
+
```
trollixx: No unnecessary whitespace changes please.

src/libs/ui/mainwindow.h
```diff
@@ -78,10 +79,13 @@ public slots:
 private slots:
     void applySettings();
     void openDocset(const QModelIndex &index);
+    void openAndFocusDocset(const QModelIndex &index);
```
trollixx: Let's instead of adding a new method, change `openDocset` to

```cpp
void openDocset(const QModelIndex &index, bool focusWebView = true);
```

src/libs/ui/mainwindow.cpp
```diff
@@ -372,9 +375,6 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
             return;
 
         openDocset(index);
```
trollixx: With the change proposed below this should be `openDocset(index, false);`

src/libs/ui/mainwindow.cpp
```diff
@@ -314,9 +314,12 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
 
     createTab();
 
-    connect(ui->treeView, &QTreeView::clicked, this, &MainWindow::openDocset);
```
trollixx: No need to change anything here with the changes I suggest below.

stecman:
Thanks, I'll sort this out.

The only change I kind of disagree with is removing `openAndFocusDocset`, but I'll follow your recommendation. What  you're suggesting is less code, but it seems pretty ugly to rely on (implicitly) missing parameters in signals to drive a UI behaviour. Plus the meaning of boolean literal arguments isn't obvious when reading the code.

trollixx:
> What you're suggesting is less code, but it seems pretty ugly to rely on (implicitly) missing parameters in signals to drive a UI behaviour.

I agree is not perfect, but `MainWindow` is already pretty messy, so I am trying to keep new changes as localized as possible, at least until I move most of the code out of one file.

> Plus the meaning of boolean literal arguments isn't obvious when reading the code.

If only we had named parameters in C++...

src/libs/ui/mainwindow.h
stecman: Hmm, looks like this isn't going to work actually - Qt allows a slot to ignore parameters provided by a signal, but it doesn't let you do the opposite:

```cpp
// mainwindow.h
void openDocset(const QModelIndex &index, bool focusWebView = true);
```

```cpp
// mainwindow.cpp
connect(ui->treeView, &QTreeView::clicked, this, &MainWindow::openDocset);
```

```
error: static assertion failed: The slot requires more arguments than the signal provides.
         Q_STATIC_ASSERT_X(int(SignalType::ArgumentCount) >= int(SlotType::ArgumentCount),
```

Doesn't work with an override of `void openDocset(const QModelIndex &index)` either, or explicitly setting signatures using `SLOT()`/`SIGNAL()` :thinking:

trollixx: Indeed, then there's no escape from two slots :)

src/libs/ui/mainwindow.cpp
stecman: I'll still inline this one though

stecman:
Updated to inline the `docsetSelectionChanged` slot :+1:

trollixx:
It's been a while, will look over this shortly, and merge for 0.7.0.

trollixx:
Sounds like a regression, since without the patch selections is cleared fine.

I still want add an option to make the new behavior opt-in. It should also nicely work with the keyboard navigation proposed in  #1007.

stecman:
I've been running a pretty old version - just rebased against master and this issue is no longer present :+1: 

Do you want to sort out that preference yourself, or should I add it here?

trollixx:
I'll add the settings, no worries, just sure where exactly to put them yet...

trollixx:
I have (finally!) implemented this functionality on top of the current code base, appears to work pretty well. Unless somebody requests a configuration option, I don't see any harm in having this always enabled.

Thank you for contributing this in the first place, and sorry for my insanely slow processing time!

PR #763, opened 2017-08-23
click on docset fills it's name to searchbox

Yanpas:
peek 2017-08-23 17-48

### DIFF
--- a/src/libs/ui/mainwindow.cpp
+++ b/src/libs/ui/mainwindow.cpp
@@ -287,7 +287,7 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     setupSearchBoxCompletions();
     SearchItemDelegate *delegate = new SearchItemDelegate(ui->treeView);
     delegate->setDecorationRoles({Registry::ItemDataRole::DocsetIconRole, Qt::DecorationRole});
-    connect(ui->lineEdit, &QLineEdit::textChanged, [delegate](const QString &text) {
+    connect(ui->lineEdit, &QLineEdit::textEdited, [delegate](const QString &text) {
         delegate->setHighlight(Registry::SearchQuery::fromString(text).query());
     });
     ui->treeView->setItemDelegate(delegate);
@@ -372,7 +372,7 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
         setupSearchBoxCompletions();
     });
 
-    connect(ui->lineEdit, &QLineEdit::textChanged, [this](const QString &text) {
+    connect(ui->lineEdit, &QLineEdit::textEdited, [this](const QString &text) {
         if (text == currentTabState()->searchQuery)
             return;
 
@@ -784,6 +784,7 @@ void MainWindow::keyPressEvent(QKeyEvent *keyEvent)
     case Qt::Key_Escape:
         ui->lineEdit->setFocus();
         ui->lineEdit->clearQuery();
+        emit ui->lineEdit->textEdited(ui->lineEdit->text()); // text is not changed programmatically
         break;
     case Qt::Key_Question:
         ui->lineEdit->setFocus();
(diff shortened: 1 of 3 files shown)


trollixx:
I really like the idea!

At the same time changing user's input on a simple click feels confusing and intrusive. I thing there should be a modifier required, e.g. <kbd>Alt</kbd>+<kbd>Click</kbd>. I am suggesting <kbd>Alt</kbd>, because <kbd>Ctrl</kbd> and <kbd>Shift</kbd> are usually used for open in a new tab or window.

Yanpas:
@trollixx Alt key is usually used by window manager in linux and won't be transmitted to custom application.
As other options we may:
* Check if input is empty or ends with `:` and constains only one `:`
* Add search icon to the right of each item
* Add right click menu with one item
* Double-click

I was testing this today and I found that it is unnecessary to change the behaviour IMO. If there is anything in the searchbar then user won't see docset list, else user has nothing to lose

Also I've [looked up](https://youtu.be/dvMzthdRA7k?t=447) at Dash, it prioritizes docset which was selected. Unintuitive but smart

trollixx:
Interesting observation about Linux WMs. I always make WM use Windows/Meta key instead.

Depending on user's DE double click can be used instead of single clicks to activate a tree view item, so not really an option. Perhaps, again a modifier + double click? A bit messy though.

Adding a search icon similar to what Dash have should be fine, although I have no idea how often people would use that feature.

> If there is anything in the searchbar then user won't see docset list, else user has nothing to lose

But once you put something in the search input, the docset tree will be replaced with search results.

Yanpas:
> observation about Linux WMs

Linux Mint and Ubuntu use  <kbd>Alt</kbd> by default and that is more than half Linux users.  <kbd>Meta</kbd> more frequently used for kind-off start menus.
> But once you put something in the search input, the docset tree will be replaced with search results.

So do you agree with my solution? Else I will implement search icon.

stecman:
What about double-click triggering this? Keyboard modifiers are a bit hidden, and as mentioned there's not really one that is obviously meaningful for this behaviour (as in, something a user would guess to press to achieve this behaviour). Double click means open on most platforms, so that makes the most sense to me.

trollixx:
> So do you agree with my solution? Else I will implement search icon.

My comment was actually against that proposal. I don't think we should switch to in-docset search when user merely clicks on one. It should be a more explicit action.

A search icon near each docset (the Dash way) feels a bit of an overkill. Perhaps, we can only show it when user hovers a docset?

Yanpas:
> we can only show it when user hovers a docset
I like this approach. I'll open new PR

Yanpas:
BTW, in `mainwindow.cpp:288` is `delegate` object leaking memory?

Yanpas:
@Not really leaking
I guess there is nothing important in the destructor

Yanpas:
It seems to me that this feature would require a lot of code to write: QTreeView has "entered" signal but lacks "left" signal. And there are no access to underlying object of QTreeView. The way I see it will be  inheriting from QAbstractItemView and reimplement a lot of logic. It's odd that Qt does not allow performing such simple tasks

trollixx:
No need to touch `QTreeView`. In `SearchItemDelegate::paint()` you can simply check for `option.state && QStyle::State_MouseOver`.

Yanpas:
And how to hide search icon? How to call paint when the mouse left this area?
We may hide upper and lower icons in the tree when the mouse hovers some another icon, but there would be always one icon visible.

We may also track vieportEntered signal and other widgets, but IMO it's ugly

trollixx:
My bad, automatically typed `&&`.  It's better to use `option.state.testFlag(QStyle::State_MouseOver)` to avoid wasting time on debugging later.

Yanpas:
Moved to #768

PR #768, opened 2017-09-03
search icon completion

Yanpas:
peek 2017-09-03 18-18
Fallback search.svg belongs to KDE Breeze theme.

### DIFF
--- a/src/libs/ui/mainwindow.cpp
+++ b/src/libs/ui/mainwindow.cpp
@@ -287,7 +287,7 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     setupSearchBoxCompletions();
     SearchItemDelegate *delegate = new SearchItemDelegate(ui->treeView);
     delegate->setDecorationRoles({Registry::ItemDataRole::DocsetIconRole, Qt::DecorationRole});
-    connect(ui->lineEdit, &QLineEdit::textChanged, [delegate](const QString &text) {
+    connect(ui->lineEdit, &QLineEdit::textEdited, [delegate](const QString &text) {
         delegate->setHighlight(Registry::SearchQuery::fromString(text).query());
     });
     ui->treeView->setItemDelegate(delegate);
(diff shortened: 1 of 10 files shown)


Yanpas:
Will commit check for parents in click slot

Yanpas:
And still I think it's a wrong way to use delegate. Maybe we should stick to `setIndexWidget`?

trollixx:
I've done some digging, and it seems like the right solution would be catching mouse events in reimplemented [QAbstractItemDelegate::editorEvent](http://doc.qt.io/qt-5/qabstractitemdelegate.html#editorEvent).

Here's the debug output I get by just logging `event` parameter:

```
QMouseEvent(MouseButtonPress, LeftButton, localPos=202,7, screenPos=990,431)
QMouseEvent(MouseButtonRelease, LeftButton, localPos=202,7, screenPos=990,431)
QMouseEvent(MouseButtonPress, LeftButton, localPos=200,10, screenPos=988,434)
QMouseEvent(MouseButtonRelease, LeftButton, localPos=200,10, screenPos=988,434)
```

trollixx:
@Yanpas Are you still interested in getting this in? I think the `editorEvent` is the way to go.

Yanpas:
Yep. I've forgot of opened PR, I will send updates this week I think

src/libs/ui/mainwindow.cpp
```diff
@@ -285,22 +285,24 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
...
-    ui->tocListView->setItemDelegate(new SearchItemDelegate(ui->tocListView));
+    ui->tocListView->setItemDelegate(new SearchItemDelegate(*ui->tocListView, ui->tocListView));
     connect(ui->tocSplitter, &QSplitter::splitterMoved, this, [this]() {
         m_settings->tocSplitterState = ui->tocSplitter->saveState();
     });
 
     createTab();
 
-    connect(ui->treeView, &QTreeView::clicked, this, &MainWindow::openDocset);
-    connect(ui->tocListView, &QListView::clicked, this, &MainWindow::openDocset);
```
Yanpas: The focus from search icon was cancelled by this actions

Yanpas:
I've updated the branch

trollixx:
Wow, ten files changed for this tiny feature. I think implementation can be leaner. I'll go over the change in a bit in more detail. Thanks!

PR #783, opened 2017-09-30
[ADD] "Open File" in File menu, open html file to new tab directly.

fantasyearl:
Hello, 

I love Zeal, it's good while searching API documentation.

But, recently, I used wget to download a entire website.

And I want to open local html files on Zeal, but looks like this function is not implement yet.

So, I try to implement it. 

I don't know if this kind of functions are necessary to Zeal, but I just want to share my result.

Thanks

### DIFF
--- a/src/libs/core/settings.cpp
+++ b/src/libs/core/settings.cpp
@@ -74,6 +74,7 @@ void Settings::load()
 
     settings->beginGroup(GroupTabs);
     openNewTabAfterActive = settings->value(QStringLiteral("open_new_tab_after_active"), false).toBool();
+    tabsMovable = settings->value(QStringLiteral("tabs_movable"), false).toBool();
     settings->endGroup();
 
     settings->beginGroup(GroupSearch);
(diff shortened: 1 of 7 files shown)


fantasyearl:
Hi, 
I made second commit.
I think movable tabs could be more convenience.
And, I add a new setting on the Preference -> Tabs.
So user can choose if they want tabs are movable.
Thanks.

trollixx:
I don't want to let users open plain HTML files. Then the next request would be adding an address bar, and eventually we might end up with another browser :)

There are tools like [dashing](https://github.com/technosophos/dashing) that can generate a docset from a set of HTML files.

Regarding your second change, I don't see why make movable tabs optional. It could be just the default non-configurable behavior. If you want, submit your second commit without settings as a separate PR.

P.S. Please do not mix unrelated changes in one pull request.

fantasyearl:
OK, I'll submit another PR with movable tabs.
Thanks for your quick reply.

PR #803, opened 2017-10-20
ui: Create new tab on tabbar doubleclick

palinek:
Just add other way to create new tab -> double click on the empty area of tab bar.


palinek:
ping

trollixx:
Merged, thanks!

PR #818, opened 2017-11-15
Small changes to fix build on CentOS 7

rogual:
On CentOS 7, `<QtConcurrent>` is not available; it needs to be `<QtConcurrent/QtConcurrent>`. This is already used in `docsetregistry.cpp`; only `filemanager.cpp` needed changing.

I also updated the readme to mention the sqlite dependency.


trollixx:
It should actually build with <QtConcurrent>. Are you using CMake or qmake?

trollixx:
And please, please, please don't put unrelated commits in a single pull-request!

rogual:
qmake && make. Happy to takeout the README change.

trollixx:
Could you try with CMake? I'll update the instructions as qmake support will be dropped after 0.5.0 release.

trollixx:
I have just fixed building with qmake, and updated the README to mention CMake. Thanks for bringing my attention to the problem

PR #852, opened 2017-12-25
Skip whitespace at the start of file

nishanthkarthik:
This is a deviation from the `XML` spec that no white space must be present before the first `xml` tag

### DIFF
--- a/src/libs/util/plist.cpp
+++ b/src/libs/util/plist.cpp
@@ -40,6 +40,11 @@ bool Plist::read(const QString &fileName)
         m_hasError = true;
         return false;
     }
+    
+    char buf;
+    // Skip whitespace at the beginning of file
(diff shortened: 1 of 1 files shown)


trollixx:
I don't see why this is needed. The plist file has to be a valid XML.

nishanthkarthik:
What is the default behavior of such malformed xml in the Dash app? Why I ask is because this docset is hosted on kapeli's servers.

trollixx:
I am quite reluctant about merging this change. I really don't like when something is not following the standard. It's the first occurrence of this problem in years, let's wait until we discover more.

Regardless, thank you for reporting the problem, and contributing the solution!

nishanthkarthik:
On second thought, I agree with you. XML is a widely used standard and hacks aren't necessary in most cases.

PR #883, opened 2018-02-09
Add a CTRL+L shortcut to bring focus to the search field

deets:
This PR adds the ability to press Ctrl-L and go to the search box (lineEdit in mainwindow.ui)

This is an established shortcut with browser ("location bar") and also used in Dash.

### DIFF
--- a/src/libs/ui/mainwindow.cpp
+++ b/src/libs/ui/mainwindow.cpp
@@ -307,6 +307,12 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     });
     ui->treeView->setItemDelegate(delegate);
 
+    QAction* goToLineEditAction = new QAction(ui->lineEdit);
+    goToLineEditAction->setShortcut(QKeySequence(Qt::CTRL + Qt::Key_L));
+    goToLineEditAction->setShortcutContext(Qt::ApplicationShortcut);
+    ui->lineEdit->addAction(goToLineEditAction);
+    connect(goToLineEditAction, &QAction::triggered, this, [this]() { ui->lineEdit->setFocus(); });
+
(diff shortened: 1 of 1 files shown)


trollixx:
There's an open issue about this #401 , but I am not convinced this is the correct approach.

Zeal doesn't have a search bar, but I'd like to add a breadcrumb-like navigation in the future. Then <kbd>Ctrl+L</kbd> would be expected to focus it.

For the current search field we already support [<kbd>Ctrl+K</kbd>](https://github.com/zealdocs/zeal/blob/acc1fdbd11f0984f16f2fd3d70ad1430bdd7c924/src/libs/ui/mainwindow.cpp#L159).

deets:
@trollixx thanks for letting me know about Ctrl-K. It works. At the same time, I can only re-iterate the comment made on the issue you mentioned - Ctrl-L is a standard. It's what I used coming from dash and all browsers, and IMHO that's a strong precedence. There is power in relying on defaults, as Ctrl-C, Ctrl-X, Ctrl-V show.

A second point then is the question of discoverability. I looked for something like that, and even if I'm in favor of Ctrl-L would have probably settled for Ctrl-K. But not learning about it is of course unfortunate. I'm writing this while using Firefox under Linux. I have to admit I also didn't find a place where Ctrl-L is mentioned. Yet it exists and works.

So I wonder if it would be worth adding a "goto search field" entry as a menu? Granted it won't be used much because it's kinda silly to use a menu with the mouse that does something you can just do by clicking on the field to begin with. But when somebody searches for a function, the menus and their depiction of short-cuts is a natural point of interaction. 

And last but not least: maybe this can be configurable? That would also open an avenue for discovery, if there is a "shortcuts" tab in the preferences.

trollixx:
Added with a bit less code in 83a8bb75fe1c4eb90ba7fc078344fefaf4feb8a2.

PR #905, opened 2018-03-12
add support for SOCKS5 proxies

### LINKED ISSUE #893: Support SOCKS5 proxy
Can zeal support socks5 proxy?

Only http proxy found.

Unpublished:
closes #893 

I don't know how to easily update these hint coords, maybe someone can update them in a followup commit.

### DIFF
--- a/src/libs/core/application.cpp
+++ b/src/libs/core/application.cpp
@@ -217,20 +217,25 @@ void Application::applySettings()
         QNetworkProxyFactory::setUseSystemConfiguration(true);
         break;
 
-    case Core::Settings::ProxyType::UserDefined: {
-        QNetworkProxy proxy(QNetworkProxy::HttpProxy, m_settings->proxyHost, m_settings->proxyPort);
-        if (m_settings->proxyAuthenticate) {
-            proxy.setUser(m_settings->proxyUserName);
-            proxy.setPassword(m_settings->proxyPassword);
-        }
-
-        QNetworkProxy::setApplicationProxy(proxy);
-
-        // Force NM to pick up changes.
-        m_networkManager->clearAccessCache();
+    case Core::Settings::ProxyType::HTTP:
+        applyProxy(QNetworkProxy::HttpProxy);
+        break;
+    case Core::Settings::ProxyType::SOCKS5:
+        applyProxy(QNetworkProxy::Socks5Proxy);
         break;
     }
+}
+
+void Application::applyProxy(const QNetworkProxy::ProxyType &type) {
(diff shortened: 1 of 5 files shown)


trollixx:
Thanks for the contribution!

Unless I missed something, I don't see any differences between HTTP and SOCKS5 code branches. I think it'd be better to have just one user defined setting, that accepts both types of proxies. Updating the preference dialog with "HTTP or SOCKS5" label should work just fine.

src/libs/core/application.cpp
```diff
@@ -217,20 +217,25 @@ void Application::applySettings()
...
-
-        QNetworkProxy::setApplicationProxy(proxy);
-
-        // Force NM to pick up changes.
-        m_networkManager->clearAccessCache();
+    case Core::Settings::ProxyType::HTTP:
+        applyProxy(QNetworkProxy::HttpProxy);
+        break;
+    case Core::Settings::ProxyType::SOCKS5:
+        applyProxy(QNetworkProxy::Socks5Proxy);
```
Unpublished: I agree with you, but here the proxy type is different.

trollixx: Indeed, I didn't notice this tiny difference...

trollixx:
I wonder if there are any valid setups when independent settings for HTTP and SOCKS5 proxies would be required..

Unpublished:
Yeah that one was easy to oversee and I believe the only one.

AFAIK one can't setup up two proxies at the same time with qt so the only usecase I can think of would be a user who switches between http/socks proxy very often (for whatever reasons) and therefore wants them to be saved separately.

trollixx:
It is possible with [`QNetworkProxyFactory`](https://doc.qt.io/qt-5/qnetworkproxyfactory.html). See a possible workflow [here](https://doc.qt.io/qt-5/qnetworkaccessmanager.html#setProxyFactory).

Unpublished:
Sorry, I'm always getting seg faults when I try to use QNetworkProxyFactory::setApplicationProxyFactory, can you finish this PR as you would like it to be?

trollixx:
I'd like to check if I can make `QNetworkProxyFactory` work, if not will just merge your change.

trollixx:
This will go into 0.7 branch.

grooveadelic:
How far off from a merge is this PR? I'm willing to give a hand in getting these feature merged, even if I'm not a C developer myself.

trollixx:
I added SOCKS5 support in 4d019fff1f21d110beb9c3bea34b775fb9760929. It's pretty much similar to this PR, except that I implemented UI a bit differently. Regardless, thanks for the contribution!

PR #911, opened 2018-03-19
assets: Add appstream metadata

barthalion:
Renamed desktop launcher to reverse DNS, which is also application ID in newly added appdata file. Release version and date (via `RELEASE_DATE` variable) is taken from top-level CMakeLists file.


trollixx:
Can we keep Flathub metadata in a separate repo? Similar to all other kinds of packaging metadata.

barthalion:
Sure, although it's not Flatpak specific. Regular distributions also make use of that file to display application details in GNOME Software/KDE Discover so it improves visibility either way.

assets/CMakeLists.txt
```diff
@@ -4,14 +4,18 @@ if(UNIX AND NOT APPLE)
     include(KDEInstallDirs)
 
     foreach(_i 16 24 32 64 128)
-        install(FILES "freedesktop/appicons/${_i}/zeal.png"
+        install(FILES "freedesktop/appicons/${_i}/org.zealdocs.Zeal.png"
```
trollixx: AFAIU this will break all icon themes, no? Plus Zeal itself tries to load icon `zeal` from the current theme.

assets/CMakeLists.txt
```diff
@@ -4,14 +4,18 @@ if(UNIX AND NOT APPLE)
...
             DESTINATION "${KDE_INSTALL_ICONDIR}/hicolor/${_i}x${_i}/apps"
             # TODO: Use `RENAME zeal.png` and get rid of subdirectories.
         )
     endforeach()
 
-    # TODO: Generate via zeal.desktop.in.
-    install(FILES "freedesktop/zeal.desktop"
+    # TODO: Generate via org.zealdocs.Zeal.desktop.in.
+    install(FILES "freedesktop/org.zealdocs.Zeal.desktop"
```
trollixx: Is this compatible with all DEs?

assets/freedesktop/org.zealdocs.Zeal.appdata.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop">
+  <id>io.zealdocs.Zeal.desktop</id>
```
trollixx: I think this should start with `org.`

assets/freedesktop/org.zealdocs.Zeal.appdata.xml
```diff
@@ -0,0 +1,48 @@
...
+  <url type="homepage">https://zealdocs.org/</url>
+  <url type="bugtracker">https://github.com/zealdocs/zeal/issues</url>
+  <url type="help">https://zealdocs.org/usage.html</url>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://i.imgur.com/qBkZduS.png</image>
+    </screenshot>
+  </screenshots>
+  <provides>
+    <id>zeal.desktop</id>
```
trollixx: This file would be absent, wouldn't it?

trollixx:
Didn't know that other `appdata.xml` may be used by other apps, then I guess we can add it to the main repo...

Additional note, I only use squash mergers, so it's better to keep one commit per PR. Sadly there's no way on GH to enforce that.

assets/freedesktop/org.zealdocs.Zeal.appdata.xml
barthalion: Oops! I'll fix it.

assets/CMakeLists.txt
barthalion: Yes, DEs generally don't care about filename.

assets/CMakeLists.txt
barthalion: It shouldn't, but I can rename it on flatpak side as well (especially if it's also in the code). I will revert this chunk after work.

assets/freedesktop/org.zealdocs.Zeal.appdata.xml
barthalion: In case of distributions shipping older release of Zeal with custom appdata, it will cause Software/Discover to merge both and prefer the latest one.

barthalion:
> Additional note, I only use squash mergers, so it's better to keep one commit per PR. Sadly there's no way on GH to enforce that.

That's fine, I will resolve all comments and squash my changes.

assets/CMakeLists.txt
trollixx: If it's some a standard/recommendation, I think we can switch, but I haven't seen other apps using such icon names.

assets/freedesktop/org.zealdocs.Zeal.appdata.xml
```diff
@@ -0,0 +1,48 @@
...
+  </provides>
+  <releases>
+    <release date="2018-02-18" version="0.6.0">
```
trollixx: Including changelog will require writing it before tagging a release. Something I tried to avoid in the past.

Do package managers show this to user when a new version is available?

barthalion: I think changelog itself isn't important, but current version is being shown in application store/flatpak CLI. Not a hard requirement yet for upstream-shipped appdata. I suppose I could make it a proper template?

barthalion: Hum, not sure how to proceed with this. Would having additional variable in top-level Cmakelists to hold release date be acceptable to you?

richarson: @trollixx Some package manager do show the changelog, Discover for example:
https://pointieststick.wordpress.com/2018/02/23/this-week-in-discover-part-7/

But it's not a hard requierement, just a nice-to-have :)

trollixx: In the perfect world release notes should be generated from somewhere, but that feels like an overengineering in Zeal's case. Whatever is the best option according to you guys, let's do that. I didn't know about this appdata thing up until a few days back.

The only thing with 0.6.0 is that it's already tagged, so I guess the next real lease will be 0.6.1.

trollixx:
Ping! :)

Btw, what's the difference between a metadata maintained by the upstream project vs flathub.org?

assets/CMakeLists.txt
trollixx: I've been trying to find any specification telling to use reverse URL naming, but failed so far.

Here's what [GNOME docs](https://developer.gnome.org/icon-naming-spec/#guidelines) say:
> An application must either use a generic application icon name provided by this specification, or install an icon named the same as the executable for running the application.

Any other sources? I kind of like the URL-based convention, but I have never seen any project using it.

richarson:
@trollixx I'm not sure what you mean by metadata, but if it's the appdata.xml file, maintained upstream can be used by flathub, snap and native package managers (and their GUIs) as is, whitout maintainig their own.

assets/freedesktop/CMakeLists.txt
```diff
@@ -12,8 +12,16 @@ if(UNIX AND NOT APPLE)
...
     )
 
-    # TODO: Generate via zeal.desktop.in.
-    install(FILES "zeal.desktop"
+    configure_file( 
+        org.zealdocs.Zeal.appdata.xml.in 
+        org.zealdocs.Zeal.appdata.xml
+    )
+
+    install(FILES ${CMAKE_BINARY_DIR}/assets/freedesktop/org.zealdocs.Zeal.appdata.xml
```
barthalion: Is there some magic variable with this path?

abn:
Similar to my comment on the appdata file name, it might be better to leave the names as is now as these changes impact non flatpak packaging.

I'd suggest we let flatpak manifest deal with this. As an example; https://github.com/abn/zeal-packaging/blob/flatpak/flatpak/org.zealdocs.Zeal.json#L7

abn:
@barthalion I have created #933 cleaning up changes a bit and remove the breaking changes; I have left the attribution intact. Hope that is okay. 

Do you have a manifest pr open on flathub yet? If not I can create one with the manifest from https://github.com/abn/zeal-packaging/tree/flatpak/flatpak.

barthalion:
@abn Let's make things correctly, unless there is so eminent comet going to hit Earth soon. Just because flatpak-builder can handle renames, doesn't mean it should be used instead of upstream changes to how files are named in the first place.

Writing manifest is really the simplest thing to accomplish here. I've had it ready when I opened this pull request and in the meantime introduced base application with qtwebkit so build times are sane. I really insist on doing the rename and appdata addition here (which will be Flathub requirement sooner or later).

trollixx:
Sorry folks, I am super limited on time for the next few weeks. Will try to review as soon as I get a moment.

matthiasclasen:
Would be great to get this landed. I've seen some interest in making zeal available on flathub. Merging this would help

trollixx:
I'll merge this soon. I've been on a  limited internet access for the past couple months.

matthiasclasen:
Thanks!

alunux:
Any update about this PR?

assets/freedesktop/CMakeLists.txt
trollixx: I am not aware of such...

trollixx:
Merged, sorry for the long wait.

PR #929, opened 2018-04-23
add support for SOCKS5 proxies, add support protocol config;

fuyb:
1. add support for SOCKS5 proxies;
2. add support protocol config;


trollixx:
There's already #905, have you seen it?

I won't be around for the next couple of weeks, so I will review once I am back.

trollixx:
SOCKS5 support has been added, see https://github.com/zealdocs/zeal/pull/905#issuecomment-436133579 for details.

Thanks for the contribution anyway!

fuyb:
The config file different than #905.
[proxy]
authenticate=false
host=127.0.0.1
password=
port=1080
protocol=1  # This 
type=2 # And This
username=

PR #951, opened 2018-06-25
Background fix for dark mode

htower:
If HTML page doesn't set background ("Bash" docset for example) then it will be theme-dependent. When light theme is used with disabled dark mode everything works well - we have black text on white background. When light theme is used with enabled dark mode everything is fine again - we have light text on dark background. But when dark theme is used with disabled dark mode we will see black text on dark background (tested on Qt 5.11). Therefore when our dark mode CSS is applied we will see light text on light background :( So we should always set default background color to white - it will be properly processed both for light and dark themes.

OLD version:
screenshot_20180625_202236

NEW version:
screenshot_20180625_202401

### DIFF
--- a/src/libs/ui/mainwindow.cpp
+++ b/src/libs/ui/mainwindow.cpp
@@ -879,7 +879,7 @@ void MainWindow::applySettings()
         removeTrayIcon();
 
     // Content
-    QByteArray ba;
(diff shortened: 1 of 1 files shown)


trollixx:
Thanks for contributing! I am not sure though that it's a good idea to override background by default.

Issue #917, opened 2018-03-28
New Tab Button


strawberry-choco:
Thoughts about adding a new tab button to the right side of the tabs? Like in browsers.

PR #968, opened 2018-08-10
Added newTab button

nishanthkarthik:
This is a follow-up to #917 
I have implemented the new tab button.


trollixx:
Thanks for your contribution! I've been quite short on time lately, but I'll try to take a look at this soon.

trollixx:
Thanks for contributing this (long time ago!), but I'd prefer a more flexible implementation, where the new tab button would stay visible, when tabs reach the right edge. I have some ideas on how to better implement that, which is quite different from this change set, hence closing the PR.

PR #1002, opened 2018-10-02
Complete TODO : Show more user friendly labels

nishanthkarthik:
This PR completes the task at https://github.com/zealdocs/zeal/blob/3bdb8ee0fd5b9062cdcba1d38aad2e09d2757bdd/src/libs/ui/docsetsdialog.cpp#L521

### DIFF
--- a/src/libs/util/readableinterval.cpp
+++ b/src/libs/util/readableinterval.cpp
@@ -1,0 +1,80 @@
+/****************************************************************************
+**
+** Copyright (C) 2015-2018 Oleg Shparber
+** Copyright (C) 2013-2014 Jerzy Kozera
+** Contact: https://go.zealdocs.org/l/contact
+**
+** This file is part of Zeal.
+**
+** Zeal is free software: you can redistribute it and/or modify
+** it under the terms of the GNU General Public License as published by
+** the Free Software Foundation, either version 3 of the License, or
+** (at your option) any later version.
+**
+** Zeal is distributed in the hope that it will be useful,
+** but WITHOUT ANY WARRANTY; without even the implied warranty of
+** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+** GNU General Public License for more details.
+**
+** You should have received a copy of the GNU General Public License
+** along with Zeal. If not, see <https://www.gnu.org/licenses/>.
+**
+****************************************************************************/
+
+#include "readableinterval.h"
+
+#include <QStringBuilder>
+#include <QStringList>
+
+using namespace Zeal::Util;
+
+ReadableInterval::ReadableInterval(QDateTime timestamp, QDateTime reference) :
+    m_timestamp(timestamp),
+    m_reference(reference)
+{
+    m_delta = m_reference.toSecsSinceEpoch() - m_timestamp.toSecsSinceEpoch();
+}
+
+ReadableInterval::ReadableInterval(QDateTime timestamp) :
+    m_timestamp(timestamp)
+{
+    m_reference = QDateTime::currentDateTime();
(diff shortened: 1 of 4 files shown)


trollixx:
Linux build failed because [`QDateTime::toSecsSinceEpoch()`](http://doc.qt.io/qt-5/qdatetime.html#toSecsSinceEpoch) was added in Qt 5.8. I'll bump the requirement for 0.7.x.

nishanthkarthik:
I'll try replacing it with a backwards compatible method later today

trollixx:
No need, the plan is to switch to newer Qt in 0.7 anyway.

src/libs/util/readableinterval.h
```diff
@@ -0,0 +1,68 @@
...
+namespace Zeal {
+namespace Util {
+
+class ReadableInterval
+{
+public:
+    ReadableInterval(QDateTime timestamp, QDateTime reference);
+    ReadableInterval(QDateTime timestamp);
+
+    QString toReadableString();
```
trollixx: Maybe just make the whole thing a static method?

nishanthkarthik:
@trollixx Done (I also fixed some clang tidy warnings). Please let me know if there's anything else :)

nishanthkarthik:
Hey, on the three LGTM warnings of comparing with known value at compile time, how do I transform this snippet to a compile time branch?

```
const int X = 3;

int a = 0;
if (m && ++a < X) 
    f();
if (n && ++a<X)
    g();
```

Here, the increment occurs only if `m` or `n` is `true`. If not, this has to go to the next branch until `a = X`. I am not sure how to optimize this out in compile time as we have no idea if first condition is true or not until runtime.

I think this is a false alert since we depend on the boolean expression short-circuiting to increment the counter.

src/libs/util/readableinterval.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+namespace {
+const qint16 SECONDSPERMINUTE = 60;
+const qint16 SECONDSPERHOUR = SECONDSPERMINUTE * 60;
+const qint32 SECONDSPERDAY = SECONDSPERHOUR * 24;
+const qint32 SECONDSPERYEAR = SECONDSPERDAY * 365;
+
+const qint8 MAX_FIELDS_DISPLAYED = 3;
+
+const QString ZERO_INTERVAL_STRING = "now";
```
trollixx: Also, it's better to avoid making translatable strings constant. Just put them inside `tr()` inline.

trollixx:
I see rightfully LGTM complains about always true conditions. Maybe rewrite the whole thing as a loop?

src/libs/util/readableinterval.cpp
nishanthkarthik: I'm leaving plural forms of units with a `(s)`. Once we apply translation to this project, qt should automatically detect the quantity and append `s`.

http://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

trollixx:
I am still thinking on how to implement this in a cleaner way. The current proposal is a bit too heavy for the task it accomplishes. Perhaps, worth checking if we can use `std::chrono` for duration manipulation...

nishanthkarthik:
Hey @trollixx 

`std::chrono::duration` looks like a viable alternative on the computation part where we split seconds to multiple duration units. It doesn't look like there is a way in this lib to print it in human readable format.

If you tell me which part of my snippet seems heavy, I can try replacing the same with `chrono` equivalent.

trollixx:
Right, we still need to print generate the text ourselves, but I think `std::chrono` should be used for manipulating the duration. That should reduce the amount of code required as well.

nishanthkarthik:
Sure, let me update this PR with chrono

Issue #999, opened 2018-09-29
High CPU usage with stutter on scroll in Qt5 documentation


nishanthkarthik:
I have jerky scrolling (both smooth scroll on and off) with qt documentation (the qdatetime page), about 80% CPU. With JavaScript disabled on the QWebView, the CPU usage drops to ~10%.

I am not able to identify which piece of js bundled with the documentation is responsible for this.

Adding an option for disabling JavaScript in QWebView looks like a reasonable workaround here.

PR #1003, opened 2018-10-06
Enable JavaScript only when Info.plist specifies isJavaScriptEnabled

nishanthkarthik:
Follow up to https://github.com/zealdocs/zeal/issues/999

Implement `isJavaScriptEnabled` through `Info.plist`


trollixx:
Merged, thank you!

PR #1004, opened 2018-10-06
Investigate TODO: QAction triggered signal doesn't connect with QToolButton animateClick slot

nishanthkarthik:
`QAction::triggered` has the prototype `triggered(bool checked=false)`
`QAbstractButton::animateClick` has the prototype `nimateClick(int msec=100)`

In this case, we don't set the QAction checkable boolean and its by default, `false`.
An implicit conversion of `false` to `int` occurs here. The actual signal to slot connection is also confirmed using GammaRay


trollixx:
Thanks!

Issue #712, opened 2017-03-31
option to cancel individual docset downloads


diveshuttam:
While downloading multiple docsets if cancel is pressed, all the current downloads get cancelled. Add an option to cancel individual downloads.

PR #1027, opened 2018-11-04
Add cancel button for each of the ongoing docset downloads

nishanthkarthik:
**Preview**

cancelbuttonspecific

The `Cancel` button cancels the current download by emitting a `cancelButtonClicked` signal. This signal is emitted by overriding `QStyledItemDelegate::editorEvent` and checking if the `QMouseEvent` coordinates fall inside the cancel button bounds

### DIFF
--- a/src/libs/ui/docsetsdialog.cpp
+++ b/src/libs/ui/docsetsdialog.cpp
@@ -67,6 +67,7 @@ constexpr int CacheTimeout = 24 * 60 * 60 * 1000; // 24 hours in microseconds
 const char DocsetNameProperty[] = "docsetName";
 const char DownloadTypeProperty[] = "downloadType";
 const char DownloadPreviousReceived[] = "downloadPreviousReceived";
+const char DownloadTotalSize[] = "downloadTotalSize";
 const char ListItemIndexProperty[] = "listItem";
 }
 
@@ -437,11 +438,12 @@ void DocsetsDialog::downloadProgress(qint64 received, qint64 total)
 
     qint64 previousReceived = 0;
     const QVariant previousReceivedVariant = reply->property(DownloadPreviousReceived);
-    if (!previousReceivedVariant.isValid())
+    if (!previousReceivedVariant.isValid()) {
+        reply->setProperty(DownloadTotalSize, total);
         m_combinedTotal += total;
-    else
+    } else {
         previousReceived = previousReceivedVariant.toLongLong();
-
+    }
     m_combinedReceived += received - previousReceived;
     reply->setProperty(DownloadPreviousReceived, received);
 
@@ -572,6 +574,8 @@ void DocsetsDialog::setupAvailableDocsetsTab()
     using Registry::DocsetRegistry;
 
     ui->availableDocsetList->setItemDelegate(new ProgressItemDelegate(this));
+    ProgressItemDelegate* del = static_cast<ProgressItemDelegate*>(ui->availableDocsetList->itemDelegate());
+    connect(del, &ProgressItemDelegate::cancelButtonClicked, this, &DocsetsDialog::cancelDownload);
 
     connect(m_docsetRegistry, &DocsetRegistry::docsetUnloaded, this, [this](const QString name) {
         QListWidgetItem *item = findDocsetListItem(name);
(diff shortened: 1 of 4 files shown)


nishanthkarthik:
This fixes https://github.com/zealdocs/zeal/issues/712

trollixx:
Thanks for the contribution!

After a quick test I've noticed that the Cancel button stays visible during installation stage, but does not do anything.

There are two options here: a) hide the button, b) make cancellation work for installation. The second requires more work, and should be submitted as a separate PR first.

src/libs/ui/docsetsdialog.cpp
```diff
@@ -572,6 +574,8 @@ void DocsetsDialog::setupAvailableDocsetsTab()
     using Registry::DocsetRegistry;
 
     ui->availableDocsetList->setItemDelegate(new ProgressItemDelegate(this));
+    ProgressItemDelegate* del = static_cast<ProgressItemDelegate*>(ui->availableDocsetList->itemDelegate());
```
trollixx: Just create delegate first, and then set it, then the'd be no need in casting.

Also coding style: `ProgressItemDelegate *delegate`.

src/libs/ui/docsetsdialog.cpp
```diff
@@ -702,6 +706,29 @@ QNetworkReply *DocsetsDialog::download(const QUrl &url)
     return reply;
 }
 
+void DocsetsDialog::cancelDownload(const QModelIndex &index)
+{
+    // Find and delete download jobs corresponding to index
+    for (QNetworkReply *reply : m_replies) {
+        if (reply->property(ListItemIndexProperty).toInt() == index.row()
```
trollixx: It's better to return as early as possible, and avoid large nested blocks. In this case simple invert the condition and and use `continue`.

src/libs/ui/progressitemdelegate.cpp
```diff
@@ -24,7 +24,10 @@
 #include "progressitemdelegate.h"
 
 #include <QPainter>
+#include <QEvent>
```
trollixx: Please sort includes alphabetically.

src/libs/ui/progressitemdelegate.cpp
```diff
@@ -69,7 +72,28 @@ void ProgressItemDelegate::paint(QPainter *painter, const QStyleOptionViewItem &
...
+    buttonRenderer->render(painter);
     painter->restore();
 
     QStyledItemDelegate::paint(painter, styleOption, index);
 }
+
+bool ProgressItemDelegate::editorEvent(QEvent *event, QAbstractItemModel *model,
+                                       const QStyleOptionViewItem &option, const QModelIndex &index)
+{
+    QMouseEvent* mouseEvent = static_cast<QMouseEvent*>(event);
```
trollixx:
```suggestion
    QMouseEvent *mouseEvent = static_cast<QMouseEvent *>(event);
```

src/libs/ui/progressitemdelegate.cpp
```diff
@@ -69,7 +72,28 @@ void ProgressItemDelegate::paint(QPainter *painter, const QStyleOptionViewItem &
...
+    QRect cancelBounds = option.rect;
+    cancelBounds.setLeft(cancelBounds.right() - cancelButtonWidth);
+
+    if (event->type() == QEvent::MouseButtonRelease
+            && index.model()->data(index, ShowProgressRole).toBool()
+            && cancelBounds.contains(mouseEvent->pos()))
```
trollixx: Curly braces are required now (I know that not all code is conformant yet).

src/libs/ui/progressitemdelegate.h
```diff
@@ -44,8 +44,15 @@ class ProgressItemDelegate : public QStyledItemDelegate
     void paint(QPainter *painter, const QStyleOptionViewItem &option,
                const QModelIndex &index) const override;
 
+    bool editorEvent(QEvent *event, QAbstractItemModel *model,
```
trollixx: `editorEvent()` is a protected method, and should never become public.

src/libs/ui/progressitemdelegate.cpp
nishanthkarthik: Now I understand the implication of this style. Thanks for pointing it out.

`int* a, b` is not int *a, *b`

trollixx:
It's been a while, and sorry for ignoring this for so long. I believe this should go in, and can be polished later. The base branch has changed a lot, and some files do not even exist any more. Rebasing is pretty much impossible, but copied changes seem to work as expected. Let me know if you want to update the PR yourself, otherwise I'll push my version (with attribution) in a couple of days.

trollixx:
I've [force-pushed](https://github.com/zealdocs/zeal/compare/2a3b4428fdeffb68a5ceaddf790547ea54a9ef93..72c73b0fad2bb532ea6170421df05d6c24f7307a) the original changes without the insanity of merge commits.

PR #1035, opened 2018-11-11
Closes #1033: focus search bar text selects all even if focused

### LINKED ISSUE #1033: "Select search bar" hotkey within search bar should select all text
ctrl+L will focus the search bar and select all the text, like a browser. While the search bar has focus, in a browser ctrl+L will select all the text (i.e. the behavior doesn't change depending on what's focused) while in zeal ctrl+L will do nothing.

I propose we make ctrl+L select all the text no matter what is focused (particularly solving the issue where ctrl+L does nothing while the search bar is focused).

mcomella:
#1033.

I'm not very familiar with C++ so I apologize for any basic mistakes. :grimacing:

### DIFF
--- a/src/libs/ui/mainwindow.cpp
+++ b/src/libs/ui/mainwindow.cpp
@@ -159,13 +159,14 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
 
     // Setup application wide shortcuts.
     // Focus search bar.
-    QShortcut *shortcut = new QShortcut(QStringLiteral("Ctrl+K"), this);
-    connect(shortcut, &QShortcut::activated,
-            ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
-
-    shortcut = new QShortcut(QStringLiteral("Ctrl+L"), this);
-    connect(shortcut, &QShortcut::activated,
-            ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
+    QShortcut *shortcut;
+    QStringList searchBarFocusShortcuts;
+    searchBarFocusShortcuts << "Ctrl+K" << "Ctrl+L";
+    for (auto it = searchBarFocusShortcuts.begin(); it != searchBarFocusShortcuts.end(); ++it) {
+        shortcut = new QShortcut(*it, this);
+        connect(shortcut, &QShortcut::activated,
+                ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
+    }
 
     // Duplicate current tab.
     shortcut = new QShortcut(QStringLiteral("Ctrl+Alt+T"), this);
--- a/src/libs/ui/mainwindow.cpp
+++ b/src/libs/ui/mainwindow.cpp
@@ -165,7 +165,7 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
     for (auto it = searchBarFocusShortcuts.begin(); it != searchBarFocusShortcuts.end(); ++it) {
         shortcut = new QShortcut(*it, this);
         connect(shortcut, &QShortcut::activated,
-                ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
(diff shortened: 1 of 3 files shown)


mcomella:
Build passes on ubuntu but fails on Windows: the error is:
```
C:\projects\zeal\src\libs\ui\mainwindow.cpp(163): error C2079: 'searchBarFocusShortcuts' uses undefined class 'std::array<QString,2>' [C:\projects\zeal\dist\src\libs\ui\Ui.vcxproj]
C:\projects\zeal\src\libs\ui\mainwindow.cpp(163): error C2440: 'initializing': cannot convert from 'initializer list' to 'int' [C:\projects\zeal\dist\src\libs\ui\Ui.vcxproj]
  C:\projects\zeal\src\libs\ui\mainwindow.cpp(163): note: The initializer contains too many elements
C:\projects\zeal\src\libs\ui\mainwindow.cpp(164): error C2228: left of '.begin' must have class/struct/union [C:\projects\zeal\dist\src\libs\ui\Ui.vcxproj]
  C:\projects\zeal\src\libs\ui\mainwindow.cpp(164): note: type is 'int'
C:\projects\zeal\src\libs\ui\mainwindow.cpp(164): error C3536: 'it': cannot be used before it is initialized [C:\projects\zeal\dist\src\libs\ui\Ui.vcxproj]
C:\projects\zeal\src\libs\ui\mainwindow.cpp(164): error C2228: left of '.end' must have class/struct/union [C:\projects\zeal\dist\src\libs\ui\Ui.vcxproj]
  C:\projects\zeal\src\libs\ui\mainwindow.cpp(164): note: type is 'int'
C:\projects\zeal\src\libs\ui\mainwindow.cpp(165): error C2100: illegal indirection [C:\projects\zeal\dist\src\libs\ui\Ui.vcxproj]
```

My assumption is that QString is not imported on Windows despite it being so on Linux so I'll explicitly add the QString header. While I'm at it, I realized I could deduplicate `QStringLiteral` by using `QString::fromStdString` in the loop. I think this allocates memory on the heap (it returns a copy) but I think that's fine because this string should live for the lifetime of the application.

mcomella:
I got the same error: I assume it's because `<array>` was not imported. That being said, the stdlib isn't used often in the codebase so I switched to QStringList which might play nicer with Windows.

src/libs/ui/mainwindow.cpp
```diff
@@ -159,13 +159,14 @@ MainWindow::MainWindow(Core::Application *app, QWidget *parent) :
...
-    connect(shortcut, &QShortcut::activated,
-            ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
-
-    shortcut = new QShortcut(QStringLiteral("Ctrl+L"), this);
-    connect(shortcut, &QShortcut::activated,
-            ui->lineEdit, static_cast<void (SearchEdit::*)()>(&SearchEdit::setFocus));
+    QShortcut *shortcut;
+    QStringList searchBarFocusShortcuts;
+    searchBarFocusShortcuts << "Ctrl+K" << "Ctrl+L";
+    for (auto it = searchBarFocusShortcuts.begin(); it != searchBarFocusShortcuts.end(); ++it) {
```
nishanthkarthik:
```suggestion
    for (auto it = searchBarFocusShortcuts.cbegin(); it != searchBarFocusShortcuts.cend(); ++it) {
```

Since we aren't modifying `it`, using `cbegin` instead of `begin` seems better.

mcomella: Makes sense to me, thanks!

trollixx:
Thanks for contributing, and I apologize for how long it took me to get to this. I was working on a major refactoring (#1081), and kind of ignored everything else.

While reviewing your change, and reading Qt docs, I discovered a better way to implement #1033 with the help of `QEvent::ShortcutOverride` events. I pushed 5ef5111bd9577ecb6eb66000542b7e563d38d6e8, and now the search box can handle `Ctrl+K`/`Ctrl+L` on its own, and set the selection accordingly. I am closing this pull request, but thank you for bringing my attention to the issue, and please test the new behavior.

PR #1037, opened 2018-11-17
Trim feed URL before trying to resolve it

vbrandl:
Sometimes, when copying a feed URL, I accidentally include a leading
whitespace. Zeal then denies the URL since it does not start with a
valid protocol indicator.

This PR trims the feed URL before using it. This might improve the UX
for some users.


trollixx:
Thanks!

Issue #428, opened 2015-11-17
Search: "xsl:if" does not find results, "xslt:if" finds "xsl:if"


cweiske:
I've installed the XSLT docset and tried to search for `xsl:if` - no results found. Quoting does not help (`"xsl:if"`).

When searching for `xslt:if`, the page `xsl:if` was found.

I'd expect that `xsl:if`already finds the `xsl:if` page.

trollixx:
Nice catch. I need to figure out a way to distinguish docset search prefixes from symbol names that contain a colon.

ssokolow:
Quoting seems like a good idea... possibly combined with an "If no results, treat it all as a literal string and re-search" fallback.

cweiske:
Thank you!

PR #1038, opened 2018-11-18
Adding support to filter by type

bmpotter13:
This pull request adds support to filter returned results by their type. The search syntax has been modified to support this. 

To filter results to only one docset, the word is prefixed by `d:`, for example `d:ruby`, to only search the Ruby docset. To filter by type, the word is prefixed by `t:`, for example `t:class`, to only return classes. As normal, any word is searched across the available docsets. These can be combined as well. For example, `d:ruby t:function inject` will search for `inject` in the Ruby docset limiting results to only those that have the type Function.

Currently, only one docset and type can be specified.


trollixx:
Hi, thanks for contributing! Sorry for taking so long to reply.

This is an interesting concept for how to narrow down the search results! I am not sure about syntax though. I can see this breaking when some docset would contain symbols with `d:` or `t:` in their names. We already have a similar problem with the XSL docset (#428). I'd like to consider other possible approaches, that would be unlikely to collide with anything.

Also we should consider how to ingrate this with plugin API, so IDEs can take full advantage of the type information.

trollixx:
Per my previous comment, I do not feel comfortable merging these implementation. Such changes require a thorough design stage. If you are still interested in this, please create an issue with the proposed feature, so we could discuss the details, and narrow down corner cases. Thank you!

Issue #170, opened 2014-06-30
Show Dash user contrib docsets in Zeal


jboadas:
Just trying to see the user contrib docsets but can't see it on the list

wosc:
Thanks @josephfinlayson for pointing out that "feed" can also be a Tarball!

However, most Tarballs are removed from the [User Contributions](https://github.com/Kapeli/Dash-User-Contributions) repository and uploaded to the @Kapeli CDN, which is not publicly accessible AFAICT, so what you need to do is trawl through the [Pull Requests](https://github.com/Kapeli/Dash-User-Contributions/pulls) and look for the Tarballs there.

This is a bit of annoying manual work, but I don't know what a more convenient solution could look like, since I don't think it would be fair to ask @Kapeli to offer these docsets for us non-paying Zeal users via bandwidth/resources that @Kapeli is paying for.

Kapeli:
Sorry for not seeing this issue earlier!

Actually I'd be glad if Zeal would use these docsets as well. I already allow access to the other (main) docsets, so I don't see a reason why not to do the same with the user contributed ones.

You can get a json of all the available user contributed docsets at http://sanfrancisco.kapeli.com/feeds/zzz/user_contributed/build/index.json

Once you have that, you can get to the docset archives like this:

```
http://<mirror>.kapeli.com/feeds/zzz/user_contributed/build/<key name inside json>/<archive name>
```

Examples:
```
http://newyork.kapeli.com/feeds/zzz/user_contributed/build/Atom/Atom.tgz
http://london.kapeli.com/feeds/zzz/user_contributed/build/SDL/org.libsdl.sdl20.docset.tgz
```

The icons in the json are base64 encoded.

Regarding the `<mirror>` part, you're supposed to choose the closest/best mirror. I've given details about how to do that to @trollixx.

trollixx:
Thank you, @Kapeli!

I am working on a revamped docset manager for Zeal, which will include user contributed docsets.

lowjoel:
Would RDoc/Yard documentation support (like Dash) belong to this ticket also?

oddityoverseer13:
:+1: to rdoc support in zeal.

NoAnyLove:
Since I was looking for docset for matplotlib, it's very happy to find this and finally soveld my problem. Thank you @Kapeli !

lowjoel:
[...]

PR #1049, opened 2018-12-07
Add User contributed docset

troopa81:
Add the capacity to see and download Dash user contributed Docset (ticket #170).

I try my best to keep the way docset download was managed (cache for instance) and to follow the coding rules (editorconfig). 

Regards,
Julien

### DIFF
--- a/src/libs/ui/docsetsdialog.cpp
+++ b/src/libs/ui/docsetsdialog.cpp
@@ -59,6 +59,13 @@ const char ApiServerUrl[] = "http://api.zealdocs.org/v1";
 const char RedirectServerUrl[] = "https://go.zealdocs.org";
 // TODO: Each source plugin should have its own cache
 const char DocsetListCacheFileName[] = "com.kapeli.json";
+const char UserContributedDocsetListCacheFileName[] = "com.kapeli.json.user_contributed";
+
+// TODO manage mirror
+const char UserContributedApiUrl[] =
+  "http://london.kapeli.com/feeds/zzz/user_contributed/build/index.json";
+const char UserContributedDocsetListUrl[] =
+  "http://london.kapeli.com/feeds/zzz/user_contributed/build/%1/%2";
 
 // TODO: Make the timeout period configurable
 constexpr int CacheTimeout = 24 * 60 * 60 * 1000; // 24 hours in microseconds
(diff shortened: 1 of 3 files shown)


src/libs/ui/docsetsdialog.cpp
```diff
@@ -59,6 +59,13 @@ const char ApiServerUrl[] = "http://api.zealdocs.org/v1";
 const char RedirectServerUrl[] = "https://go.zealdocs.org";
 // TODO: Each source plugin should have its own cache
 const char DocsetListCacheFileName[] = "com.kapeli.json";
+const char UserContributedDocsetListCacheFileName[] = "com.kapeli.json.user_contributed";
+
+// TODO manage mirror
+const char UserContributedApiUrl[] =
+  "http://london.kapeli.com/feeds/zzz/user_contributed/build/index.json";
+const char UserContributedDocsetListUrl[] =
+  "http://london.kapeli.com/feeds/zzz/user_contributed/build/%1/%2";
```
sighingnow: Maybe we should use `http://kapeli.com/feeds/zzz/user_contributed/build/index.json` and `http://kapeli.com/feeds/zzz/user_contributed/build/%1/%2`. for better CDN usage.

Kapeli: "kapeli.com" just goes to the server that hosts my website, which is in the same datacenter as newyork.kapeli.com. I prefer if you don't use kapeli.com but one of the other mirrors.

sighingnow: I see, I had expected "kapeli.com" has geo-based DNS solver, pointing to mirrors in other regions. Thanks for the clarification.

troopa81:
Just a gently reminder for this PR to be merged

trollixx:
Thanks for the contribution, and sorry for the delayed response, lately I haven't had much time for Zeal.

I don't really want to merge this, because my idea is that Zeal should only talk to `api.zealdocs.org`. That gives much more control, and should allow unified approach for handling all docsets on the client side.

I hoped to make changes on the server-side rather soon, but if that will not happen in time for the next release, I'll merge this PR as a temporary solution.

troopa81:
I understand your will to manage this on server side.

Zeal is a great tool and the only thing preventing me from using it more widely was the access to user docset. 

I keep an eye on the project waiting for server side modifications so. If you need to merge this PR for temporary solution and need some modifications, just let me know.

ryanpcmcquen:
@trollixx, any progress here?

PR #1064, opened 2019-01-11
Update .gitignore

brainplot:
Hi, I checked out the code and opened it with QtCreator; I started a build and the IDE generated a directory called `build-Desktop-Debug/` which was not being ignored by Git. I took the chance to update the `.gitignore` file using the `Qt` template from [gitignore.io](https://www.gitignore.io), which fixed the issue along with adding more rules that should prevent any unwanted file from being shown as untracked in the future.

In order to check if any _wanted_ file was being left out, I ran `git check-ignore *`, which gave the following output:
```zsh
➜ git check-ignore * 
build-Desktop-Debug
CMakeLists.txt.user
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,16 +9,38 @@
 *.dll
 *.dylib
 
-# Qt Creator
+# Qt-es
+object_script.*.Release
+object_script.*.Debug
+*_plugin_import.cpp
+/.qmake.cache
+/.qmake.stash
+*.pro.user
+*.pro.user.*
+*.qbs.user
+*.qbs.user.*
+*.moc
+moc_*.cpp
+moc_*.h
+qrc_*.cpp
+ui_*.h
+*.qmlc
+*.jsc
+Makefile*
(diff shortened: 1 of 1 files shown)


trollixx:
In-source builds are unsupported, hence the situation should never happen. Qt Creator by default creates build directories on the same level as the one with the sources, so you might have accidentally changed that. 

Other additions are unnecessary, since Zeal doesn't use QML, QMake, qbs, etc.

brainplot:
Oh, I must have missed that in-source buids are unsupported. Yes, I changed the default build directory because I find that build outputs clutter up my project folder. Anyway, thanks for the info!

wnear:
Can I manually copy CMakeList.txt to CMakeList.txt.user, do some modifications,  then offer latter file as an option to cmake cli for build so that I don't need to change the original CMakeList.txt and the git repo stays clean? 
 I checked the cmake cli manual and online, nothing related found.

PR #1065, opened 2019-01-11
Replace Q_OS_OSX with Q_OS_MACOS

brainplot:
[According to the Qt docs](https://doc.qt.io/qt-5/qtglobal.html#Q_OS_OSX), the `Q_OS_OSX` macro is deprecated and
`Q_OS_MACOS` should be used instead.

I ran a mixture of `find` and `sed` over the `src/` folder to replace all occurrences.

Not a big deal after all, but I thought it was a fair change.


trollixx:
Merged, thanks!

PR #1068, opened 2019-01-15
Avoid creation of unnecessary objects

brainplot:
Qt containers return a deep copy when used in a loop with a call to a member function such as [`QMap::keys()`](https://doc.qt.io/qt-5/qmap.html#keys) or [`QMap::values(const Key& key)`](https://doc.qt.io/qt-5/qmap.html#values-1).
Using iterators or assigning the return value to a const object avoids the copy, thanks to the implicit sharing feature of Qt.
See the code examples [here](https://doc.qt.io/qt-5/qtglobal.html#qAsConst).

Further, when arguments in a constructor are to be passed to a member's constructor, taking the argument by value and then moving it can be beneficial when the argument is an rvalue as it will avoid the call to a copy constructor, thanks to move semantics.
See [here](https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html).

### DIFF
--- a/src/libs/registry/docset.cpp
+++ b/src/libs/registry/docset.cpp
@@ -41,6 +41,7 @@
 #include <sqlite3.h>
 
 #include <cstring>
+#include <utility>
 
 using namespace Zeal::Registry;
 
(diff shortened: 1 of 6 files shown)


src/libs/registry/docset.cpp
```diff
@@ -428,8 +429,11 @@ void Docset::countSymbols()
 // TODO: Fetch and cache only portions of symbols
 void Docset::loadSymbols(const QString &symbolType) const
 {
-    for (const QString &symbol : m_symbolStrings.values(symbolType))
-        loadSymbols(symbolType, symbol);
+    // itPair is a QPair<QMap::const_iterator, QMap::const_iterator>, with itPair.first and itPair.second respectively
```
trollixx: Nice, but comment is really unnecessary, because it just explains what is already clear from the documentation for `equal_range`.

PR #1080, opened 2019-02-09
Set default zoom of webview based on DPI.

yiding:
Control zoom factor directly rather than using hardcoded zoom levels.

In order to ensure that the webview starts out scaled appropriately to the
current font DPI, we set default zoom level based on current DPI, e.g.:

  96 DPI - 100% zoom
  120 DPI - 125% zoom

Since this will allow zoom levels beyond the predefined set, I also change the
logic for zooming to simply add or remove 10% to the current scaling.

Note that if we want to still maintain the predefined zoom levels, I can rewrite this to multiply the predefined zoom levels by the zoom level derived from the current DPI.

### DIFF
--- a/src/libs/browser/webview.cpp
+++ b/src/libs/browser/webview.cpp
@@ -39,7 +39,7 @@
 using namespace Zeal::Browser;
 
 WebView::WebView(QWidget *parent)
-    : QWebView(parent), m_zoomLevel(0)
+    : QWebView(parent)
 {
     page()->setNetworkAccessManager(Core::Application::instance()->networkManager());
     // Applies any DPI based scaling.
@@ -63,7 +63,7 @@ void WebView::setZoomLevel(int level)
     m_zoomLevel = level;
 
     // Scale the webview relative to the DPI of the screen.
-    qreal dpiZoomFactor = logicalDpiY() / 96.0;
+    const float dpiZoomFactor = logicalDpiY() / 96.0;
 
     setZoomFactor(availableZoomLevels().at(level) / 100.0 * dpiZoomFactor);
     emit zoomLevelChanged();
--- a/src/libs/browser/webview.cpp
+++ b/src/libs/browser/webview.cpp
@@ -39,9 +39,11 @@
 using namespace Zeal::Browser;
 
 WebView::WebView(QWidget *parent)
-    : QWebView(parent)
+    : QWebView(parent), m_zoomLevel(0)
 {
     page()->setNetworkAccessManager(Core::Application::instance()->networkManager());
+    // Applies any DPI based scaling.
+    setZoomLevel(defaultZoomLevel());
 }
 
 int WebView::zoomLevel() const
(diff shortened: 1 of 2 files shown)


trollixx:
Thank you for the contribution, very interesting! Does this work better on High-DPI screens? What about moving window between screens with different DPI or scaling factors?

The list of zoom levels is inspired by what Firefox and Chrome do. Also maintaining specific zoom values should make it easier to save and restore sessions in the future.

yiding:
> Does this work better on High-DPI screens?

This works better for high dpi screens where the user is using font dpi scaling (vs full screen scaling). Font DPI scaling usually looks better than full screen scaling when scaling by fractional amounts. For full screen scaling (e.g. scaling by 200% using the display control panel), this shouldn't change anything. I'm only talking about in the context of windows, I'm not sure whether windows use DPI based scaling or not (I should probably check).

> What about moving window between screens with different DPI or scaling factors?

I'm not familiar with how QT surfaces this information, and I've never had much hope that this works correctly on linux. I think this is somewhat orthogonal to this diff though.

> The list of zoom levels is inspired by what Firefox and Chrome do. Also maintaining specific zoom values should make it easier to save and restore sessions in the future.

I can do that.

yiding:
Alright I've updated this to apply the DPI based scaling with the zoom-level style scaling. Seems to behave sensibly on linux in KDE5 and Gnome3 when font DPI scaling is used, and when screen-scaling is used.

trollixx:
@yiding thanks, I'll review and test over the weekend.

src/libs/browser/webview.h
```diff
@@ -41,7 +41,7 @@ class WebView : public QWebView
     void setZoomLevel(int level);
 
     static const QVector<int> &availableZoomLevels();
-    static const int &defaultZoomLevel();
+    static int defaultZoomLevel();
```
trollixx: Good catch.

src/libs/browser/webview.cpp
```diff
@@ -60,7 +62,10 @@ void WebView::setZoomLevel(int level)
 
     m_zoomLevel = level;
 
-    setZoomFactor(availableZoomLevels().at(level) / 100.0);
+    // Scale the webview relative to the DPI of the screen.
+    qreal dpiZoomFactor = logicalDpiY() / 96.0;
```
trollixx: `qreal` is deprecated, please use `float`. Also should be `const`.

src/libs/browser/webview.cpp
```diff
@@ -39,9 +39,11 @@
 using namespace Zeal::Browser;
 
 WebView::WebView(QWidget *parent)
-    : QWebView(parent)
+    : QWebView(parent), m_zoomLevel(0)
```
trollixx: Just set it to zero in the header.

trollixx:
@yiding could you rebase on top of master?

trollixx:
Thank you!

PR #1085, opened 2019-03-09
libs: enforce static linking of internal libs

henning-schild:
Since the application binary is the only file installed and the libs are
not used by anyone else, make sure they are always static. Otherwise
cmake can make them shared and those shared libs will be missing after
install.

Signed-off-by: Henning Schild <henning@hennsch.de>

### DIFF
--- a/src/libs/browser/CMakeLists.txt
+++ b/src/libs/browser/CMakeLists.txt
@@ -1,4 +1,4 @@
-add_library(Browser
+add_library(Browser STATIC
     webcontrol.cpp
     searchtoolbar.cpp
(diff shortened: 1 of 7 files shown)


trollixx:
Interesting, I've never seen the libs being linked dynamically, when could that happen?

henning-schild:
I saw it happening on a gentoo Linux working on a new package. That workflow uses cmake/ninja possibly with some flags that cause it. If you need a more profound justification i can probably reproduce that.

trollixx:
Merged, thanks!

PR #1103, opened 2019-05-11
Two minor refactorings and change of representation of docset updatability

mfurkanuslu:
05e4018 is for avoiding code repeat.
2ac6445: introduced change in this commit: installed docset list shows updatable items in bold instead of showing extra text for that.

### DIFF
--- a/src/libs/registry/listmodel.cpp
+++ b/src/libs/registry/listmodel.cpp
@@ -100,13 +100,18 @@ QVariant ListModel::data(const QModelIndex &index, int role) const
             return QVariant();
         }
     case ItemDataRole::DocsetNameRole:
-        if (index.parent().isValid())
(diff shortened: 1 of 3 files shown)


mfurkanuslu:
AppVeyor failure isn't related to my commits. Jobs in CI with Ubuntu1804 image are broken.

trollixx:
Thank you for the contribution, but I do not believe the changes would be beneficial to the project. The first small refactoring would add to the cognitive complexity of already hard to read code. And the style change might make Zeal violate user's theme settings, and/or system's HIG. The current update notification in cursive already looks foreign, and requires fixing.

P.S. In the future, please first discuss UI/UX, or any significant changes by creating an issue, or via IRC/Gitter.

Issue #924, opened 2018-04-10
Feature request: premium version


latortuga:
Several other issues have been opened and closed about the ads issue so I won't beat a dead horse too hard. Advertisements don't respect your users' privacy - today Carbon Ads is an ethical ad network, tomorrow it could be sold to the highest bidder.

Please consider accepting money from your actual customers rather than ad networks. Your market is developers and I know they'd would be happy to pay you for your work. As a business owner, I happily OK and pay for any and all software requests by any of my developers.

trollixx:
Let me first clarify that the ads in Zeal are not meant to make any real money. At some point Carbon folks reached out to me, and I just said why not. It's kind of fun to see something dripping into my PayPal account each month, but frankly I don't use the funds for anything aside from paying some tiny maintenance costs.

If you are concerned about privacy, feel free to disable the ads in the settings. Some Linux distributions do that by default.

I've been considering ways to make some real profits from Zeal, but have never done anything to make that happen. I am more inclined towards providing some enterprise solutions, while keeping Zeal itself free and open source. With the current state of Zeal, I really don't see how would introducing a premium version or a subscription plan worth it.

Also, I do have a full-time job, so Zeal gets very little of my attention. Sometimes weeks pass between commits, or me simply addressing an issue ticket. I am not sure it would be fair to charge any money in such situation, or even collect donations.

trollixx:
It's kind of a chicken and egg problem. To make Zeal profitable, I'd need to spend more time on it, but for that I'd need to make it profitable first :)

Frankly, I really don't want Zeal to be a paid or a closed source app. So, I am reluctantly looking into other ways to monetize it. And, no not with ads.

In the meanwhile I'll keep this ticket open as a reference for the similar questions I get every now and then. And everyone is welcome to express their opinion.

SirCmpwn:
Throw up a Patreon and Liberapay page and I'd throw in some dough. If you just need to meet recurring costs it can be a good approach.

vpiotr:
I'd pay for Zeal to keep it on the surface. Maybe not 100's of US dollars, but a small fee per year (for updates).

Patabugen:
+1 for at least a Patreon or donations page - Zeal is one of those apps I'd really miss if it were no longer around. It helps me make money every day, and I'd like to be able to feed some of that back to the person who make my life easier.

trollixx:
Okay, for those who asked, here's the link: https://www.buymeacoffee.com/trollixx

[...]

PR #1122, opened 2019-07-27
Add text, shield, link to buymeacoffee

HeroicKatora:
As mentioned in #924, making it more visible in the Readme which was not done yet due to time constraints. The picture used for the badge is from one of the badge styles they offer for embedding into html-based forums.


trollixx:
Thanks!

Issue #811, opened 2017-11-02
Full dark theme


eromoe:
Previous issue https://github.com/zealdocs/zeal/issues/466 add `Dark Mode` which only make main window dark. But that looks incongruous with other parts, especially left sidebar looks too shinning in dark mode.

Hope can embed a full dark theme .

oblitum:
Does this make sense? I think it would require a lot of UI rewriting to achieve, maybe it's just better left as it is currently, at user option to match widgets with the system or specific style:

![](https://i.imgur.com/uIeZM0Q.png)

trollixx:
@oblitum I guess this is more of an issue for Windows users, who do not have all the flexibility Linux DEs provide.

Styling Qt apps works on Windows as well, but most people are just not aware of it. So, I think the first step would be providing some docs on how to change Zeal's appearance with third party themes for Qt.

trollixx:
I haven't tried (QDarkStyleSheet)[https://github.com/ColinDuquesnoy/QDarkStyleSheet] yet, but it looks promising.

oblitum:
Agreed!

technodrome:
Hello guys, is there a chance this is going to be a real thing? The app itself is extremely useful, just the interface is extremely bright. I know you have your own professional and private lives, just asking.

nzig:
For those still looking for a solution:
Installing [Kvantum](https://github.com/tsujan/Kvantum/tree/master/Kvantum) worked for me. Note that after installing you need to followe the [instructions](https://github.com/tsujan/Kvantum/tree/master/Kvantum) on setting it as the default.

ezxpro:
> For those still looking for a solution:
> Installing [Kvantum](https://github.com/tsujan/Kvantum/tree/master/Kvantum) worked for me. Note that after installing you need to followe the [instructions](https://github.com/tsujan/Kvantum/tree/master/Kvantum) on setting it as the default.

For those on Windows, no solution, apparently.

oblitum:
> I don't understand why make a darkmode that only applies to a certain part of the window, but anyways.

@ezxpro because people with decent OS can then couple it to their dark desktop environment.

Zeal can only do its best to provide flexibility where it can (Qt styling), not where it cannot (Window decorations), that's generally responsibility of the OS.

K8-and-a-half-1000:
> @ezxpro because people with decent OS can then couple it to their dark desktop environment.

[...]

PR #1124, opened 2019-07-31
Add dark UI theme and custom theme support

kaygdev:
Enable user CSS and dark mode for the UI from the settings dialog. Fix some graphics issue with dark mode described in #466.

For the dark mode is used the dark orange stylesheet from http://discourse.techart.online/t/release-qt-dark-orange-stylesheet/2287 with small modifications.

### DIFF
--- a/src/app/resources/css/darkorange.stylesheet
+++ b/src/app/resources/css/darkorange.stylesheet
@@ -1,0 +1,441 @@
+QToolTip
+{
+     border: 1px solid black;
+     background-color: #ffa02f;
+     padding: 1px;
+     border-radius: 3px;
+     opacity: 100;
+}
+
+QWidget
+{
+    color: #b1b1b1;
+    background-color: #323232;
+}
+
+QWidget:item:hover
+{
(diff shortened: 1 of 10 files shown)


JARVIS-AI:
Did you test it ? (Run it with your changes)

trollixx:
Bundling unmaintained styles is not a good idea. Any operating system or Qt update will potentially break it. Also it would require extensive testing on various systems with different settings.

If something in Zeal breaks custom styles, that needs to be addressed, and I'd be happy to accept such changes, but otherwise styles should not be part of the project.

There's a ticket #811 about supporting dark window style.

0xboz:
I was trying to add the setup I used (Debian 10), but it was closed by the bot. I guess I will just leave it here in case someone comes across later. 

```shell
sudo apt install qt5-style-plugins
echo "export QT_QPA_PLATFORMTHEME=gtk2" >> ~/.profile
```
Log out and log back in. Here is the result (on Debian 10 + Gnome + Flat Remix GTK )

zeal-dark-theme-debian-gnome cleaned

Issue #865, opened 2018-01-09
Touch Screen Support


trollixx:
Touch screen gestures were [reported](https://github.com/zealdocs/zeal/issues/356#issuecomment-355859629) not to work.

trollixx:
With Qt WebEngine this should work.

PR #1125, opened 2019-08-07
feat(browser,ui): port to Qt WebEngine

kaygdev:
Porting from the obsolete QWebKit to the QWebEngine. Most features should work with it (QWebEngine does not have all features like QWebKit).

In addition, this modification blocks requests from local pages to web pages. Some documentation sets track the user with Google Analytics, this is now blocked.

### DIFF
--- a/src/app/resources/browser/assets/css/darkmode.css
+++ b/src/app/resources/browser/assets/css/darkmode.css
@@ -1,11 +1,11 @@
 /* Dark mode (see #466). */
 /* TODO: Check if we need -webkit-filter for old Qt WebKit. */
+
 html {
-    -webkit-filter: invert() hue-rotate(180deg) contrast(70%);
-    filter: invert() hue-rotate(180deg) contrast(70%);
+    background-color: #262626 !important;
+    filter: invert(1) hue-rotate(180deg) contrast(70%) !important;
 }
 
 html img {
-    -webkit-filter: invert() hue-rotate(180deg) contrast(120%);
-    filter: invert() hue-rotate(180deg) contrast(120%);
+    filter: invert(1) hue-rotate(180deg) contrast(120%) !important;
 }
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,7 +7,7 @@ branches:
     - stable
     - /release\/.*/
 image:
-  - Visual Studio 2015
+  - Visual Studio 2017
 configuration: Release
 platform:
   - x86
@@ -19,9 +19,9 @@ environment:
 init:
 - ps: |-
     if ($Env:PLATFORM -eq "x64") {
-        $Env:QT_ROOT = "C:\Qt\5.11\msvc2015_64"
+        $Env:QT_ROOT = "C:\Qt\5.12\msvc2017_64"
     } else {
-        $Env:QT_ROOT = "C:\Qt\5.11\msvc2015"
+        $Env:QT_ROOT = "C:\Qt\5.12\msvc2017"
     }
 
     $Env:Path += ";${Env:QT_ROOT}\bin;C:\Program Files\7-Zip"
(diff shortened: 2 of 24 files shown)


trollixx:
I would really like adding support for Qt WebEngine, but without losing backend Qt WebKit at least as a build option. Refactoring out the `browser` component was partially motivated by the idea of having different web view engines. The work is not finished yet, and I think here we can identify the missing pieces, and add necessary abstractions to accommodate both engines.

@kaygdev Would you be interested in continuing to work on this PR to land it as an option, rather than a replacement?

I see the need for `QWeb(Engine)Settings` to be encapsulated within the component. Another to-do item would be breaking direct inheritance of `Browser::WebView` from `QWebView`. More things might come up along the way.

kaygdev:
Hi trollixx,

Sorry for the late response. I can change the replacement of QWebKit to a build option. However, this will take some time, because I currently have absolutely no time. I hope/think that this will change in spring.

trollixx:
Build time option is what Zeal used to have a while back, when QWebEngine was quite unstable. I believe it's much better now, but I don't want to drop support for Qt WebKit. Also it would be really great to use web engine provided by the platform on Windows, and that require having a proper abstraction layer.

I'll close this for now, feel free to reopen or create a new PR whenever you get time to work on this. It'd would also be better to discuss the design beforehand. Thanks!

trollixx:
After spending more time figuring out how to deal with dual backend, and testing out this PR, I think it'd be much better to ship the next version of Zeal with Qt WebEngine rather than wait on the updated Qt WebKit. Once the next QtWK is released we can reassess the situation, and add it as the second backend. I think developing and abstraction layer will also be easier with WK2 based QtWK.

trollixx:
@kaygdev I'll rebase your changes on top of the current master, and merge.

kaygdev:
Hi trollixx,

thanks for accepting and rebasing my pull request. You can contact me when a new version of Webkit is available. I hope I have time then and can support you with the implementation.

src/app/resources/browser/assets/css/darkmode.css
```diff
@@ -1,11 +1,28 @@
 /* Dark mode (see #466). */
 /* TODO: Check if we need -webkit-filter for old Qt WebKit. */
+
+body {
+    background-color: white !important;
+    transition-duration: 0s !important;
```
trollixx, line 6: @kaygdev why is this needed?

src/app/resources/browser/assets/css/darkmode.css
```diff
@@ -1,11 +1,28 @@
...
+    filter: invert(1) hue-rotate(180deg) contrast(70%) !important;
 }
 
 html img {
-    -webkit-filter: invert() hue-rotate(180deg) contrast(120%);
-    filter: invert() hue-rotate(180deg) contrast(120%);
+    filter: invert(1) hue-rotate(180deg) contrast(120%) !important;
+}
+
+::-webkit-scrollbar {
```
trollixx, line 18: I think scroll bars should follow the system theme not web page, will remove this.

src/app/resources/browser/assets/css/darkmode.css
kaygdev: As far as I remember, a Docset still had a light background and was unreadable, but I could no longer reproduce this.
The second line seems to be an artifact from when I made the changes just for me. It suppressed an annoying animation in the Qt docset (highlight bright), I found this very annoying. Apparently that doesn't happen anymore either. I think the lines can be removed.

trollixx: Thanks for the quick response!

src/app/resources/browser/assets/css/darkmode.css
kaygdev: The problem with this was, in dark mode I have a complete dark window but with bright scrollbars in the docset part. The WebEngine don't use the the scrollbar style from the Qt Application.

src/app/resources/browser/assets/css/darkmode.css
trollixx: Oh, I noticed the issue with Qt docset, probably that should be fixed in the docset, and not by us.

src/app/resources/browser/assets/css/darkmode.css
trollixx: I see. OK, let's keep this for now. In general I think the dark mode should be smarter and follow application's theme.

trollixx: Unfortunately, it's a bit broken, no buttons. I'll look into styling scroll bars later.

trollixx: image

chris-morgan: I have always steadfastly refused to use ::-webkit-scrollbar styling because it mangles the scrollbar in this way. Hopefully they’ll follow Firefox’s lead soon, in implementing scrollbar-color and the rest of that lot, *and* in defaulting to a dark scrollbar (without ruining the controls) on dark content.

asouini:
This also fixes https://github.com/zealdocs/zeal/issues/865

tested this ci build https://ci.appveyor.com/project/zealdocs/zeal/builds/32459448 on a windows surface

trollixx:
@asouini thanks for reporting! I'll link it later.

PR #1130, opened 2019-09-02
Fix missing app icon on Wayland

lightbulbjim:
Wayland uses the .desktop file to find the app icon. Without it being specified the correct icon is not shown. 

Tested on Gnome/Wayland.

References:
* https://community.kde.org/Guidelines_and_HOWTOs/Wayland_Porting_Notes#Application_Icon
* https://doc.qt.io/qt-5/qguiapplication.html#desktopFileName-prop


trollixx:
Thank you!

PR #1136, opened 2019-09-16
fix(ui): correctly position search completer label on HiDPI display

hexchain:
Before: 
image

After: 
image

Test carried out under KDE, with a scaling factor of 2.

Note that I did not really test how it behaves on a normal DPI screen or multiple screens with different DPI, because of lack of devices.


trollixx:
Thank you for the contribution!

PR #1137, opened 2019-09-25
Fix Qt 5.11 deprecation warning for QFontMetrics::width

brainplot:
Building on my Arch Linux box with Qt 5.13.1 installed showed a few warnings saying that the `QFontMetrics::width()` and `QDir::operator=(const QString&)` were deprecated. I replaced them with the new alternatives and wrapped them with a compile-time check, which wasn't needed for `QDir::operator=(const QDir&)`. Now it should compile cleanly.

Since I saw you already made a very similar fix, by skimming through the log, I reused the same commit message of that old commit.

### DIFF
--- a/src/libs/core/extractor.cpp
+++ b/src/libs/core/extractor.cpp
@@ -56,7 +56,7 @@ void Extractor::extract(const QString &sourceFile, const QString &destination, c
 
     QDir destinationDir(destination);
     if (!root.isEmpty()) {
-        destinationDir = destinationDir.filePath(root);
+        destinationDir.setPath(destinationDir.filePath(root));
     }
 
     // TODO: Do not strip root directory in archive if it equals to 'root'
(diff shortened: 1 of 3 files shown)


src/libs/core/extractor.cpp
```diff
@@ -56,7 +56,7 @@ void Extractor::extract(const QString &sourceFile, const QString &destination, c
 
     QDir destinationDir(destination);
     if (!root.isEmpty()) {
-        destinationDir = destinationDir.filePath(root);
+        destinationDir.setPath(destinationDir.filePath(root));
```
trollixx: Unrelated to the rest of the patch, what's the reason for this?

brainplot: I mentioned it in my first statement. `QDir::operator=(const QDir&)` has gone deprecated too.
I strangely [don't see this reported](https://doc.qt.io/qt-5/qdir.html#operator-eq) on the website though. Here's the compiler warning:
```
../src/libs/core/extractor.cpp: In member function ‘void Zeal::Core::Extractor::extract(const QString&, const QString&, const QString&)’:
../src/libs/core/extractor.cpp:59:54: warning: ‘QDir& QDir::operator=(const QString&)’ is deprecated: Use QDir::setPath() instead [-Wdeprecated-declarations]
   59 |         destinationDir = destinationDir.filePath(root);
      |                                                      ^
In file included from /usr/include/qt/QtCore/QDir:1,
                 from ../src/libs/core/extractor.cpp:25:
/usr/include/qt/QtCore/qdir.h:110:11: note: declared here
  110 |     QDir &operator=(const QString &path);
      |           ^~~~~~~~
```

trollixx: I didn't see the mention of `QDir::operator=(const QString&)` until now, lol. Merging, thanks!

PR #1151, opened 2019-10-31
Modernize code base with clang-tidy's modernize-*

brainplot:
I ran the code base through the `modernize-*` checks of clang-tidy.
Enhancements include the use of `auto` in order to avoid repeating the type name when obvious from the context, the use of `constexpr` for array literals and usage of move semantics to avoid unnecessary copies.

### DIFF
--- a/src/libs/browser/webview.cpp
+++ b/src/libs/browser/webview.cpp
@@ -227,7 +227,7 @@ void WebView::mouseReleaseEvent(QMouseEvent *event)
                         .arg(clickedLink.toString()));
 
 
-            QCheckBox *checkBox = new QCheckBox("Do &not ask again");
+            auto checkBox = new QCheckBox("Do &not ask again");
             mb->setCheckBox(checkBox);
 
             QPushButton *openInBrowserButton = mb->addButton(tr("Open in &Desktop Browser"),
--- a/src/libs/sidebar/proxyview.cpp
+++ b/src/libs/sidebar/proxyview.cpp
@@ -28,13 +28,15 @@
 
 #include <QVBoxLayout>
 
+#include <utility>
+
 using namespace Zeal;
 using namespace Zeal::Sidebar;
 
-ProxyView::ProxyView(ViewProvider *provider, const QString &id, QWidget *parent)
+ProxyView::ProxyView(ViewProvider *provider, QString id, QWidget *parent)
     : View(parent)
     , m_viewProvider(provider)
-    , m_viewId(id)
+    , m_viewId(std::move(id))
 {
     setLayout(WidgetUi::LayoutHelper::createBorderlessLayout<QVBoxLayout>());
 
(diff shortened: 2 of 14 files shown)


src/libs/browser/webview.cpp
```diff
@@ -227,7 +227,7 @@ void WebView::mouseReleaseEvent(QMouseEvent *event)
                         .arg(clickedLink.toString()));
 
 
-            QCheckBox *checkBox = new QCheckBox("Do &not ask again");
+            auto *checkBox = new QCheckBox("Do &not ask again");
```
trollixx: In this case there is no point in `auto *`, and it should simply be `auto`. Same applies to similar situations below.

src/libs/ui/qxtglobalshortcut/qxtglobalshortcut_p.h
```diff
@@ -79,7 +79,7 @@ class QxtGlobalShortcutPrivate : public QAbstractNativeEventFilter
     bool setShortcut(const QKeySequence &shortcut);
     bool unsetShortcut();
 
-    virtual bool nativeEventFilter(const QByteArray &eventType, void *message,
+    bool nativeEventFilter(const QByteArray &eventType, void *message,
                                    long *result) override;
```
trollixx: This line needs to be re-indented to match the change above.

src/libs/ui/settingsdialog.cpp
```diff
@@ -38,10 +38,10 @@ using namespace Zeal::WidgetUi;
 
 namespace {
 // QFontDatabase::standardSizes() lacks some sizes, like 13, which QtWK uses by default.
-const int AvailableFontSizes[] = {9, 10, 11, 12, 13, 14, 15, 16, 17, 18,
+constexpr int AvailableFontSizes[] = {9, 10, 11, 12, 13, 14, 15, 16, 17, 18,
                                   20, 22, 24, 26, 28, 30, 32, 34, 36,
```
trollixx: Broken indentation here and below.

src/libs/sidebar/proxyview.cpp
```diff
@@ -27,14 +27,15 @@
 #include <ui/widgets/layouthelper.h>
 
 #include <QVBoxLayout>
+#include <utility>
```
trollixx: Please add an empty line above.

brainplot:
That should do it. Let me know if there's any other issue!

brainplot:
It might be a good idea to create a `.clang-format` file in the project root in order to enforce a particular style across all the code base as well as indentation issues; and have `clang-format` fix all that automatically.
There are a few online interactive generators, if that might be of interest.

Thanks for merging this!

trollixx:
Good idea! Created https://github.com/zealdocs/zeal/issues/1159 for this.

PR #1152, opened 2019-10-31
Improve error message when storage does not exist

brainplot:
When the user modifies the docsets directory from the settings page, the docsets page shows `Docset storage is read-only` when the actual source of the error is that the directory does not exist on the file system. This PR fixes that by making the label show the correct error message based on the actual error.

### DIFF
--- a/src/libs/ui/docsetsdialog.cpp
+++ b/src/libs/ui/docsetsdialog.cpp
@@ -83,7 +83,9 @@ DocsetsDialog::DocsetsDialog(Core::Application *app, QWidget *parent)
     qt_ntfs_permission_lookup++;
 #endif
 
-    m_isStorageReadOnly = !QFileInfo(m_application->settings()->docsetPath).isWritable();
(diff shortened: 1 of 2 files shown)


src/libs/ui/docsetsdialog.cpp
```diff
@@ -95,7 +97,10 @@ DocsetsDialog::DocsetsDialog(Core::Application *app, QWidget *parent)
 #endif
 
     ui->statusLabel->clear(); // Clear text shown in the designer mode.
-    ui->readOnlyLabel->setVisible(m_isStorageReadOnly);
+    ui->storageStatusLabel->setVisible(m_isStorageReadOnly);
+    ui->storageStatusLabel->setText(fi.exists() ?
+                                    QStringLiteral("<b>Docset storage is read only.</b>") :
```
trollixx: FYI, user visible strings should be wrapped into [`tr()`](https://doc.qt.io/qt-5/qobject.html#tr), so they would be translatable.

Issue #1155, opened 2019-11-08
Content rendering error in MDN docsets


dania-prog:
After updating to zeal 0.6.1, the content could not be displayed in the Tabs. Pasted below  is the display error.
![image](https://user-images.githubusercontent.com/52634550/68454947-71262b00-01fa-11ea-903d-4d00eeacbb39.png)
kindly assist to resolve. Thanks

sanruby:
error same as , the javascript docset

mansur-r:
Same error, using just an initial html docset

antoni-minkiewicz:
Same error

falegh:
Zeal 0.6.0, Ubuntu 18.04

`Content rendering error` after installing updates to these docsets:
- CSS
- HTML
- JavaScript
- SVG

React, Bash and some others working fine

dania-prog:
The following docsets has the error (content rendering error) after installing updates to the docsets:

- CSS
- HTML
- JS
- Java SE11 (the application closes out)
- Java SE9 (the application closes out)
- Perl
- RubyMotion
- SVG
- WordPress
The rest are functional. Kindly lets know when it has been resolved.

trollixx:
Docsets produced from MDN (HTML, CSS, JS, SVG) are broken because of the new JS functionality. It appears to be fixed by the https://github.com/mdn/kuma/pull/5996. Unfortunately, WebKit (and also Firefox) can't load polyfills over the `file:` scheme. I'm looking into possible solutions.

@dania-prog I cannot reproduce crashes with the Java docsets. Is it possible that you are affected by https://github.com/zealdocs/zeal/issues/868?

Perl and RubyMotion docsets didn't have index page set, should be fixed now. Refresh the docset list, and reinstall the docsets.

The WordPress docset has incorrect `dashIndexFilePath` in `Info.plist`, `no_index_test/index.html` doesn't exist. @Kapeli can you take a look? I think the property is redundant because `index.html` is present.

Kapeli:
I sometimes set the `dashIndexFilePath` to inexistent pages to make Dash not show an index for the docset anymore. In a future version of Dash index pages won't be as important anymore, so some docsets won't have them anymore.

zjffun:
it can fix temporary by remove the
`CSS.docset\Contents\Resources\Documents\developer.mozilla.org\static\build\js\react-main.ddbbd86b0685.js`.

prurigro:
[...]

PR #1164, opened 2019-11-21
core,ui: Provide ability to set a custom javascript polyfills

intact:
You can use [iterators-polyfill](https://www.npmjs.com/package/iterators-polyfill) for MDN docsets (or you can get NodeList.prototype.@@iterator-polyfill here [Polyfill.io](https://polyfill.io/v3/url-builder/))

### DIFF
--- a/src/libs/browser/webview.cpp
+++ b/src/libs/browser/webview.cpp
@@ -32,6 +32,7 @@
 
 #include <QCheckBox>
 #include <QDesktopServices>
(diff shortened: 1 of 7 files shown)


trollixx:
Thank you for contributing!

While this change could be a workaround for #1155, I don't believe Zeal should become a full featured browser. I am working on upgrading our web view backend, which will resolve the problem without requiring user intervention.

PR #1182, opened 2020-01-28
feat: serve docset pages over HTTP

trollixx:


trollixx:
Pending cpp-httplib v0.5.4 release.

Issue #1207, opened 2020-03-28
Broken twitter links


pfez:
Twitter links are broken in `src/app/resources/browser/404.html` and `src/app/resources/browser/welcome.html`.
The problem is a typo: `twiter` instead of `twitter`.

PR #1208, opened 2020-03-28
Fix broken twitter links in 404.html, welcome.html

pfez:
Twitter links were broken due to a typo (`twiter` instead of `twitter`).
This pull request fixes the links.

This fixes https://github.com/zealdocs/zeal/issues/1207


trollixx:
Thanks!

PR #1219, opened 2020-04-25
Fix compilation error with Qt 5.15

data-man:


data-man:
Sorry, double-clicked on button. :(

Issue #275, opened 2015-02-16
UI translation


trollixx:
Zeal needs to be translated in some popular languages.

Here's a list of top 10 languages (according to the today's statistics of zealdocs.org visitors, excluding English):
1. French
2. Simplified Chinese
3. Russian
4. Spanish
5. German
6. Portuguese (Brazil)
7. Japanese

Potentially (require better usage analytics):
- Polish
- Dutch
- Italian

An online translation system should be set up. Current options:
- Transifex
- OneSky
- Weblate

Perhaps, the list of supported languages should be kept small to ease maintaining of updated translations.

Alkaid-Benetnash:
I want to help translate zeal's UI into Simplified Chinese. How can I help?

trollixx:
Hi, nothing has been done for translation yet, so just keep an eye on this issue. I'll post an update once there's a translation system in place.

JIghtuse:
Russian here. Ready to help when you choose a system.

Horoneru:
French here. I will help whenever you got a translation system ready !

asdrubalivan:
I will gladly help with Spanish, ping me when the translation system is in place!

trollixx:
Thank you guys, I'll try to spend some time on the translation infrastructure soon.

Horoneru:
You could try to use Qt's translation system? Then you send the file to be translated to the translator, he gives it to you, and you only need to update the translations.

trollixx:
@Horoneru that way would cause additional headache when strings change (happens quite often). A translation platform would allow us to manage all changes easily and update translations accordingly. I guess I'll setup a SkyOne account or a Weblate instance. I am just trying to minimise my future work and make stuff automatic as much as possible.

Another reason why I am not very quick on translations is the fact, that all documentation is anyway in English. So, the translated UI wouldn't help people much. In my opinion software developers have to know English nowadays. :)

TobiX:
Maybe hosted weblate? http://weblate.org/en/hosting/

Kapeli:
> Another reason why I am not very quick on translations is the fact, that all documentation is anyway in English. So, the translated UI wouldn't help people much

Fwiw, this is also why Dash is only available in English. However, I do see how some users might prefer having all apps on their system in their native language, even if they use English docs. I haven't seen a huge demand for it though, got very few requests for it.

> In my opinion software developers have to know English nowadays

[...]

PR #1255, opened 2020-07-18
增加中文支持

yunzhongyue:
界面及文档树增加中文支持（基于Qt国际化技术）


trollixx:
I assume this is support for the Chinese language. See #275 for my reasoning on why I don't want to support UI translation. At least until we have translated documentation I don't see much value in this. Adding static translations to the code base is not really maintainable and should be done via translation platform.

Please also note, that commit messages and code comments must be in English for any contribution.

As of now, I don't believe this is feasible to merge, sorry.

yunzhongyue:
@trollixx ok,I understand.

PR #1286, opened 2020-09-30
fix(assets): prevent duplicating application icons in GNOME dock

yosmoc:
Currently, Zeal has a problem with duplicating the application icon when opening the application in Linux GNOME environment.

Exact the same problem in the following link.
https://askubuntu.com/questions/975178/duplicate-application-icons-in-ubuntu-dock-upon-launch

In order to solve this problem, we need to specify `StartupWMClass=Zeal` in `org.zealdocs.Zeal.desktop`


trollixx (approved):

trollixx:
Thanks!

PR #1315, opened 2021-01-26
build(appimage): add dependencies, disable Qt WebEngine sandbox

azubieta:
Fix missing symbol on system fontconfig for AppImage package


azubieta:
@trollixx Please test the latest binaries. I had to bring quick patch the qt5 helper that's why  I'm still using master. I'll make a new release this week and send you another PR with the version fixed.

trollixx:
@azubieta the new binary works for me, thank you!

It indeed complains about missing `qtwebengine_locales`. Is it something `appimage-builder` should take care of?

trollixx (approved):

PR #1322, opened 2021-02-20
fix(browser): prevent WebEngine taking away focus from searchbox

xaizek:
Version on master doesn't let me enter search with pauses or switch among results. Browser takes the focus away from the search box each time it's updated.

Happens since 5360da3d7f34906ff41c10ac74e6b01636598903.


trollixx (approved):

trollixx:
Thanks! Sorry it took me so long, it's been a rough year.

Issue #10, opened 2013-04-23
Adding improved support for keyboard navigation


lingnand:
Zeal is now working really well for me now; it's fast, and it doesn't take up additional space. The only thing I feel a bit missing is keyboard support for navigation. I understand this would not mean much importance for such a small application, but it still makes me wish that I don't need to lift my hands off my keyboard when I'm programming.

XeCycle:
I wonder if vimperator-like keyboard navigation would be possible here.  (Personally I switched back to reading docs online in vimperator) That is going to be a large amount of work, I guess; not familiar with Qt I cannot help.  I considered creating a Firefox plugin or XULRunner app, but that would run on OS X and compete with Dash, so am afraid that may not be allowed in-app download of Kapeli hosted docsets.

If implementing that in Zeal is practical, I'd expect its usability at the level of vimium, no need to match that of vimperator.

Btw a simple feature I would love is half-screen scrolling.

trollixx:
I think the best approach is to make all keybindings configurable. Also, adding some built-in settings (e.g. VIM-like, or web browser-like) should satisfy most users out of the box.

dmocek:
+1 for support for configurable keybindings.  I only use the mouse when I have no other option.  I'd like to see keybindings for moving down and up the list.  I prefer to use CtrlN (next) and Ctrl-P (previous) instead of the up and down arrow keys.

sic-f:
One way to move from the document window back to the search results is Ctrl+K then Tab.

dmocek:
Excellent!, this works.  I would still like to change this as Ctrl-F or Ctrl-S is the typical search keyboard shortcut.

rmolin88:
I second the vim like mappings. :100: :+1:  for a great application

madand:
This would be awesome addition to the awesome tool. Here is a list of bare-minimum bindings that should make most vim users happy:

| Vi-like binding | Description | Equivalent default binding |
| --- | --- | --- |
| `j` | Scroll one line down | `Down Arrow` |
| `k` | Scroll one line up | `Up Arrow` |
| `h` | Scroll left (horizontal) | `Left Arrow` |
| `l` | Scroll right (horizontal) | `Right Arrow` |
| `CTRL-d` | Scroll page down | `PgDn` |
| `CTRL-u` | Scroll page up | `PgUp` |

areckx:
Looking through the source code:

[...]

PR #1323, opened 2021-02-20
Add Ctrl+N and Ctrl+P keys to the search box

xaizek:
Provide <kbd>Ctrl+N</kbd> and <kbd>Ctrl+P</kbd> keys as equivalents for <kbd>Down</kbd> and <kbd>Up</kbd> keys. Similar to how it works in various other search boxes, e.g. in QtCreator's magic box.

### DIFF
--- a/src/libs/ui/searchsidebar.cpp
+++ b/src/libs/ui/searchsidebar.cpp
@@ -393,7 +393,21 @@ bool SearchSidebar::eventFilter(QObject *object, QEvent *event)
         case Qt::Key_PageUp:
             QCoreApplication::sendEvent(m_treeView, event);
             break;
+        }
 
+        if (e->modifiers() == Qt::ControlModifier) {
+            switch (e->key()) {
+            case Qt::Key_N: {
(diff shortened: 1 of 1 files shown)


dcunited001:
what's the status of this pullreq?

xaizek:
You made me realize that I committed extra stuff. Thanks.

trollixx:
Zeal is generally trying to offer shortcuts similar to web browsers. <kbd>Ctrl+N</kbd> and <kbd>Ctrl+P</kbd> have specific meaning in browsers such as creati new windows and printing. Both features have been requested to be implemented in Zeal.

When we add configurable shortcuts I can see having extra binding for list navigation but until then I don't think we should make these standard.

trollixx:
I've tested this change and the behavior becomes quite confusing. All browsers intercept <kbd>Ctrl+N</kbd> and <kbd>Ctrl+P</kbd> for creating new windows and invoking print dialog. Even in Qt Creator these are used for jumping to the first/last elements of the magic box. I think if this behavior is desirable it needs to be an option and not the default. I'll close this PR. Potentially this can be tackled as part of solution for #10.

Issue #1364, opened 2021-11-10
build failed on macos :  Could NOT find LibArchive (missing: LibArchive_INCLUDE_DIR) 


yunfengsay:
![image](https://user-images.githubusercontent.com/15527378/141047052-30f54efd-baf6-4a22-9ec3-e7af9a78f78e.png)

To fix this problem:
add "set(LibArchive_INCLUDE_DIR "/usr/local/opt/libarchive/include")" to CMakeLists.txt
![image](https://user-images.githubusercontent.com/15527378/141047154-6e6b82dd-0c7f-4f0b-9915-fd6590fc3960.png)

PR #1365, opened 2021-11-10
fix: Some macos user unstandard library path cause LibArchive_INCLUDE…

yunfengsay:
https://github.com/zealdocs/zeal/issues/1364


trollixx:
Any idea why CMake cannot discover this path on its own?

stefangehn:
Just like PR #1386 I think this is a matter of telling cmake and its standard LibArchive lookup where to find a non-standard libarchive installation.

trollixx:
Okay, looks like this is an issue due to Homebrew's installation path. 

* https://github.com/Homebrew/legacy-homebrew/issues/21415

trollixx:
whoops. I messed the PR up trying to rebase, will push manually.

trollixx:
Pushed as 294800645e5f50fb7ef2f1e4dbb69658b0d64a60. Thanks!

Issue #1022, opened 2018-10-30
Support MITM certificates on Windows


michaeljmcd:
I am attempting to use Zeal (version 0.6.1 of the Windows Portable install, 0.6.0 on Ubuntu/WSL) behind a corporate proxy with a self-signed certificate. I am unable to download any docsets, receiving the "SSL handshake failed" error message each time. I presume this is due to a need to import the proxy certificates, but I am not sure how to do this.

The certificates are already imported into the system store, as evidenced by other applications working.

How do I trust the additional certificates for the corporate proxyu so that I can download docsets?

trollixx:
After some research it appears that on Windows Qt/OpenSSL do not load certificates from the system's certificate store. I also was unable to find an easy way to add a trusted certificate to Zeal without recompiling it.

We need to address this, since MITM setups are quite common in the corporate world, but I don't think I will work on this until the new app backend is out.

trollixx:
Combination of #1367 and #1521 should address this.

PR #1367, opened 2021-11-11
fix: Include system CA certificates in requests

nevesnunes:
On Fedora 34, I've built Zeal with Qt5.15.2 and OpenSSL-1.1.1l.

When trying to check for updates, the request fails with "SSL handshake error". I can also reproduce this issue in one of the [Qt examples](https://code.qt.io/cgit/qt/qtbase.git/tree/examples/network/http?h=5.15), which includes the more descriptive message "The issuer certificate of a locally looked up certificate could not be found".

The root cause seems to be the same as described in [another issue](https://github.com/owncloud/client/issues/1540): If I run `strace` on `zeal`, I confirm that it doesn't access my certificate bundle (`/etc/pki/tls/certs/ca-bundle.crt`), instead it tries to lookup `c_rehash` generated files that aren't present.

Regardless of which files are present, Zeal can be resilient to these cases and load the certificate bundle explicitly. My pull request accomplishes that by using `QSslConfiguration`'s built-in method that provides those system certificates. This way, the peer certificate chain can be verified and no SSL handshake error is thrown.

### DIFF
--- a/src/libs/core/networkaccessmanager.cpp
+++ b/src/libs/core/networkaccessmanager.cpp
@@ -71,5 +71,9 @@ QNetworkReply *NetworkAccessManager::createRequest(QNetworkAccessManager::Operat
         op = QNetworkAccessManager::GetOperation;
     }
 
+    QSslConfiguration sslConfig = overrideRequest.sslConfiguration();
+    sslConfig.setCaCertificates(QSslConfiguration::systemCaCertificates());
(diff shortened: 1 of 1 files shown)


trollixx (approved):

trollixx:
Thanks! Somehow I missed this feature completely. This might be also a fix #1022.

PR #1386, opened 2022-04-26
build(cmake): support linux system compilation - find_package(LibArchive)

goodleixiao:
support linux system compilation
 find_package(LibArchive)


goodleixiao:
the WIN can't compile, I don't think it has started to compile.

// -------------------------------------------------------------------------------------
   Error: Building package zlib:x64-windows failed with: BUILD_FAILED
  Please ensure you're using the latest portfiles with `.\vcpkg update`, then
  submit an issue at https://github.com/Microsoft/vcpkg/issues including:
    Package: zlib:x64-windows
    Vcpkg version: 2020.06.15-nohash

goodleixiao:
#1364

stefangehn:
I don't think changes like this are needed. On my Ubuntu 22.04 installation libarchive is found just fine. Also cmake itself ships with the necessary LibArchive package lookup since at least version 3.0.2: https://cmake.org/cmake/help/v3.0/module/FindLibArchive.html

If that standard cmake module fails to find your version of FindLibArchive then maybe something is off with the libarchive installation on your system.

trollixx:
#1365 offers a simpler fix.

PR #1392, opened 2022-05-03
Port to Qt6

Arthapz:


mohamad-supangat:
@barthalion hay do you have any change about this qt6 port

mohamad-supangat:
@barthalion hay do you have any change about this qt6 port? ??

trollixx:
Thanks for the PR but I believe it's vital to keep support for Qt 5. Also significant amount of changes here are whitespace only, which would not be accepted. I have a dated local branch which builds with both Qt 5 and Qt 6. I'll push it shortly.

trollixx:
I've just pushed draft as #1420.

PR #1420, opened 2022-07-23
feat: add Qt 6 support

trollixx:
Adds support for Qt 6 along with preserving option to build with Qt 5.


mohamad-supangat:
great

PR #1434, opened 2022-10-10
Feature Request: Bookmark Option

Shams-Raju:
Feature Request:
------------------
Can you please add a new page bookmark option. If pages could be bookmarked then it would be easy to access our regular Classes, Functions, Hooks etc...


trollixx:
Spam PR?

PR #1449, opened 2023-01-09
Keep dark background on welcome page when in dark mode, rather than inverting it

macsplit:
Before:

before

After:
after


macsplit:
Problem is not happening on other PCs I've tried, so maybe it can be addressed without this patch. So withdrawing the request.

PR #1450, opened 2023-01-11
chore: remove unused methods

ckerr:
This PR removes unused functions and class methods.

I ran keepassxc through [xunused](https://github.com/mgehre/xunused), manually inspected the results to remove false positives, and then removed the code that seems to actually be unused.


trollixx:
The methods in question are getters that were either used at some point or were added for future use case. I'd rather keep them all around. Cheers.

PR #1475, opened 2023-03-21
fix(ui): fix RTL layout in list delegates

### LINKED ISSUE #1463: All custom delegate layouts are broken in RTL (Right-to-left) environment
docsets (updates available):
docsets (updates available)

search highlight:
search highlight

search categories:
search categories

ratijas:
After:

image

image

image

Fixes #1463

### DIFF
--- a/src/libs/ui/docsetlistitemdelegate.cpp
+++ b/src/libs/ui/docsetlistitemdelegate.cpp
@@ -67,6 +67,7 @@ void DocsetListItemDelegate::paint(QPainter *painter,
 #else
     textRect.setLeft(textRect.right() - fontMetrics.width(text) - 2);
 #endif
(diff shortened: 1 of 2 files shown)


src/libs/ui/docsetlistitemdelegate.cpp
```diff
@@ -59,13 +59,19 @@ void DocsetListItemDelegate::paint(QPainter *painter,
     font.setItalic(true);
 
     const QFontMetrics fontMetrics(font);
+    const int margin = 4; // Random small number
```
trollixx, line 62: There's probably margin provided by theme we should use here. I'll look into this later.

trollixx (approved):
RTL is hard, thanks for working on this!

PR #1476, opened 2023-03-21
feat(browser): add Ctrl+Enter shortcut for Highlight All option

### LINKED ISSUE #1467: Find in page: Ctrl+Enter should toggle Highlight All mode
Currently, focusing the "Find in page" search field and pressing Ctrl+Enter key combination does nothing special, it's just treated as an Enter and finds next match.

#### Expected behavior

In Firefox, Ctrl+Enter in the "Find in page" field toggles Highlight All mode, which is quite convenient.

#### Version

Zeal: 0.6.1
Qt: 5.15.8

ratijas:
Fixes #1467


trollixx (approved):
Thanks!

Issue #1520, opened 2023-08-01
SSL Handshake Filed in zeal 0.6.1 for Windows 10


binrusdi:
Hi, saya mengalami permasalahan **SSL Handshake failed** ketika ingin mendownload docsets.

![gambar](https://github.com/zealdocs/zeal/assets/119654495/3d6e2e50-c155-4938-9ade-e411f4321ae4)

dongzhaolee:
I get the same err on Windows 11, still waiting for the developers to solve the issue @trollixx

Giulidan:
I have the same error, using Windows 11.

trollixx:
Please use CI builds. Binaries are attached to each build.

Giulidan:
> 

I don't know where to find them, they all appear to me as "Expired"

marckes04:
how do I solve this??

ostap-tymchenko:
Probable duplicate of: https://github.com/zealdocs/zeal/issues/1518

binrusdi:
> Please use CI builds. Binaries are attached to each build.

Thanks, I'll try it

Epic621:
CI Build Artifacts that I see are all expired

FalcoGer:
The SSL certificate for `go.zealdocs.org` has expired on 2023-07-27

DeckerSU:
https://github.com/zealdocs/zeal/issues/1307#issuecomment-1664827774 - patch for ignore SSL errors, just made it for myself :wink:

Mintonmu:
emmmm, so, how do i solve this? only build zeal with ci on windows?  Is there any other solution？

dongzhaolee:
If you guys really have no way to solve the SSL err, you can just consider downloading from the mirror website, which deployed in some specific regions. 
Also, if you need the icons, just find it by yourself and name it 'icon.png' & 'icon@2x.png'. Then put them in the docsets folders.

Although this doesn't solve your problem, at least you can get started.

dongzhaolee:
If you're really unable to solve the SSL err, maybe you should consider downloading from the mirror website, which deployed in some specific regions.
Also, if you need the icon, just find it by yourself and name it 'icon.png' & '[icon@2x.png](mailto:icon@2x.png)'. Then put them in the docsets folders.

Although this can't solve your problem, at least you can get started.In fact, except CI builds, DeckerSU's patch works.

DeckerSU:
[...]

Issue #1307, opened 2021-01-07
SSL handshake failed (0.6.1) 


xFurther:
I still get this error on Windows 10.

#1088

What can I do to fix this?

![image](https://user-images.githubusercontent.com/6121565/103887838-b277da80-511e-11eb-8cee-fdb8993bcd0f.png)

dcunited001:
I'm running into the same problems. There's a redirect from https://go.zealdocs.org/d/com.kapeli/Clojure/latest to https://mumbai.kapeli.com/feeds/Clojure.tgz, which fails on SSL handshake. I think mumbai.kapeli.com's cert has expired.

arsenypoga:
Same issue as the above. Redirect that leads to the https://mumbai.kapeli.com seems to have expired certs.

arsenypoga:
If you *really* want the working docsets, there is a workaround. [This](https://github.com/Kapeli/feeds) git repo contains the feeds for the docs. What you can do, is copy the raw URL to the feed and import it through `Add Feeds` option under `Tools > Docsets`

trollixx:
Duplicate of #1330.

raulgm00:
> En mi caso, tampoco pude descargar ninguno de los conjuntos de documentos disponibles con Zeal `v0.6.1`instalado desde los repositorios oficiales de Ubuntu 22.04 a través de `sudo apt-get install zeal`. Para intentar resolver este problema, instalé dependencias adicionales:
> 
> ```
> sudo apt-get install extra-cmake-modules libqt5x11extras5-dev libxcb-keysyms1-dev libsqlite3-dev
> ```
> 
> y construí Zeal desde este repositorio usando lo siguiente:
> 
> ```shell
> mkdir build
> cd build
> cmake -DCMAKE_BUILD_TYPE=Release ..
> cmake --build . -- -j$(nproc)
> ```
> 
> Sin embargo, el problema persistió. Luego de una mayor investigación, descubrí que el certificado SSL en https://go.zealdocs.org (RedirectServerUrl) había expirado el `Thursday, July 27, 2023 at 1:00:09 AM` . A pesar de esto, el contenido requerido todavía estaba accesible. Para evitar los errores de SSL, escribí el siguiente parche:
> 
> ```diff
> diff --git a/src/libs/ui/docsetsdialog.cpp b/src/libs/ui/docsetsdialog.cpp
> index 5195454..bfab2da 100644
> --- a/src/libs/ui/docsetsdialog.cpp
[...]

PR #1521, opened 2023-08-04
feat(core): add option to ignore SSL errors

DeckerSU:
This pull request adds network settings to ignore SSL errors in order to handle the cases described here:

- https://github.com/zealdocs/zeal/issues/1307
- https://github.com/zealdocs/zeal/issues/1520

When the SSL certificate on the server is expired or invalid, users can still download from that server by choosing to ignore SSL errors.

If the project maintainers find this feature unacceptable, I apologize.

image

### DIFF
--- a/src/libs/ui/docsetsdialog.cpp
+++ b/src/libs/ui/docsetsdialog.cpp
@@ -675,6 +675,15 @@ bool DocsetsDialog::updatesAvailable() const
 QNetworkReply *DocsetsDialog::download(const QUrl &url)
 {
     QNetworkReply *reply = m_application->download(url);
+
+    if (m_application->settings()->isIgnoreSSLErrorsEnabled) {
+        // Connect to the reply's sslErrors signal to handle SSL errors
+        connect(reply, &QNetworkReply::sslErrors, [=](const QList<QSslError>& errors){
+            // Ignore all SSL errors
+            reply->ignoreSslErrors();
+        });
+    }
+
     connect(reply, &QNetworkReply::downloadProgress, this, &DocsetsDialog::downloadProgress);
     connect(reply, &QNetworkReply::finished, this, &DocsetsDialog::downloadCompleted);
     m_replies.append(reply);
--- a/src/libs/ui/docsetsdialog.cpp
+++ b/src/libs/ui/docsetsdialog.cpp
@@ -676,14 +676,6 @@ QNetworkReply *DocsetsDialog::download(const QUrl &url)
 {
     QNetworkReply *reply = m_application->download(url);
 
-    if (m_application->settings()->isIgnoreSSLErrorsEnabled) {
-        // Connect to the reply's sslErrors signal to handle SSL errors
-        connect(reply, &QNetworkReply::sslErrors, [=](const QList<QSslError>& errors){
-            // Ignore all SSL errors
-            reply->ignoreSslErrors();
-        });
-    }
-
     connect(reply, &QNetworkReply::downloadProgress, this, &DocsetsDialog::downloadProgress);
     connect(reply, &QNetworkReply::finished, this, &DocsetsDialog::downloadCompleted);
     m_replies.append(reply);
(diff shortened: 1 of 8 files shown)


src/libs/ui/settingsdialog.ui
```diff
@@ -745,6 +745,22 @@
          </layout>
         </widget>
        </item>
+       <item>
+        <widget class="QGroupBox" name="groupBox_8">
+         <property name="title">
+          <string>Other</string>
```
trollixx, line 751:
```suggestion
          <string>Security</string>
```

src/libs/ui/settingsdialog.ui
```diff
@@ -745,6 +745,22 @@
...
+         </property>
+         <layout class="QVBoxLayout" name="verticalLayout_15">
+          <item>
+           <widget class="QCheckBox" name="ignoreSSLErrorsCheckBox">
+            <property name="text">
+             <string>Ignore SSL errors when downloading</string>
```
trollixx, line 757:
```suggestion
             <string>Ignore SSL errors</string>
```

src/libs/ui/settingsdialog.cpp
```diff
@@ -330,5 +334,12 @@ void SettingsDialog::saveSettings()
     settings->proxyUserName = ui->proxyUsernameEdit->text();
     settings->proxyPassword = ui->proxyPasswordEdit->text();
 
+    if (settings->isIgnoreSSLErrorsEnabled != ui->ignoreSSLErrorsCheckBox->isChecked()) {
+        // https://forum.qt.io/topic/125873/how-to-clear-qnetworkreply-ignoresslerrors/2
+        m_application->networkManager()->clearAccessCache();
+    }
```
trollixx, line 340: This is already done in [`Core::Application::applySettings`](https://github.com/zealdocs/zeal/blob/94247fd4d03e6ae079cae8e50e6eb9a67fc0afad/src/libs/core/application.cpp#L274-L275), but needs to be moved outside of proxy selection switch.

src/libs/ui/docsetsdialog.cpp
```diff
@@ -675,6 +675,15 @@ bool DocsetsDialog::updatesAvailable() const
...
 {
     QNetworkReply *reply = m_application->download(url);
+
+    if (m_application->settings()->isIgnoreSSLErrorsEnabled) {
+        // Connect to the reply's sslErrors signal to handle SSL errors
+        connect(reply, &QNetworkReply::sslErrors, [=](const QList<QSslError>& errors){
+            // Ignore all SSL errors
+            reply->ignoreSslErrors();
+        });
+    }
```
trollixx, line 685: This signal should be handled on [`Core::NetworkAccessManager`](https://github.com/zealdocs/zeal/blob/94247fd4d03e6ae079cae8e50e6eb9a67fc0afad/src/libs/core/networkaccessmanager.h) to cover all network activity performed by Zeal.

The easiest approach would probably be handling the signal in `Core::Application`, when network manager is [created](https://github.com/zealdocs/zeal/blob/94247fd4d03e6ae079cae8e50e6eb9a67fc0afad/src/libs/core/application.cpp#L64). Not the most elegant solution, but would avoid the need to pass settings into `Core::NetworkAccessManager`.

trollixx (changes requested):
Thanks for contributing! Even ignoring the certificate expiration problem from the last week, I think this is a good workaround for all the SSL related issues reported over the years.

I have a few minor suggestions, please see inline.

src/libs/ui/docsetsdialog.cpp
DeckerSU: You meant like this - https://github.com/zealdocs/zeal/pull/1521/commits/b4e2c86816c435e0497030d284eccc0243c32b74 ?

src/libs/ui/settingsdialog.cpp
DeckerSU: Done - https://github.com/zealdocs/zeal/pull/1521/commits/c3be7f451322005604ec6b174e5fdb413793981b .

trollixx (approved):
Thanks for the quick turnaround.

DeckerSU:
> Thanks for the quick turnaround.

Not at all, happy to assist! Thank you for providing those suggestions and explanations.

PR #1522, opened 2023-08-07
Add a screenshot taken on a Linux desktop for Linux app stores

Mailaender:
as the existing Windows 10 based one is odd on places like https://flathub.org/apps/org.zealdocs.Zeal


trollixx:
Thanks for contributing, but I don't think it's nice to directly use Linux Mint's screenshots.

It is a valid issue with packaging metadata. I'll update all screenshots for v0.7.0, and will provide a dedicated image for Linux.

PR #1560, opened 2023-10-05
added devcontainer support

niposch:
I experimented with VS Code's devcontainers, and I was pleasantly surprised. Got this project up and running and even got the gui to show up on my Windows desktop despite the application running inside a docker container. Breakpoints also work. 
image

Setup is pretty straight forward: have docker and VS Code + devcontainer extension installed. 
Have the extension set up a dev container. It will set up the environment according to what is specified in the config files, which should make it a lot easier for people to start writing code without having to worry about setting everything up correctly, as the project can set itself up inside a docker container.

The setup works very well on Windows 11 (can't comment on older versions/other operating systems) but it would allow someone who isn't quite as familiar with how to get the project compiled/get all the dependencies installed up and running.

Feel free to reject if this isn't something the project is interested in.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,31 @@
+FROM mcr.microsoft.com/devcontainers/cpp:1-ubuntu-22.04
+
+ARG REINSTALL_CMAKE_VERSION_FROM_SOURCE="3.22.2"
+
+# Optionally install the cmake for vcpkg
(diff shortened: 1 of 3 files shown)


trollixx:
I haven't used devcontainer stuff in VS Code, but it seems to be not that different from building in WSL directly. I don't think IDE specific setup should be part of the main repo. A wiki page with instructions would be a better fit I think. Thanks.

Issue #1602, opened 2024-02-25
Incorrect path to brew's libarchive on ARM macOS


alt-romes:
Homebrew's default location for ARM/M-series chips macOSs is `/opt/homebrew/opt/libarchive/include` (see /opt/homebrew/opt/libarchive/include). This causes the `cmake -B build` to fail because `LibArchive_INCLUDE_PATH` is set to the path homebrew uses for Intel macs).

These are the offending lines:
https://github.com/zealdocs/zeal/blob/de213a74d8bd45b40638dd6d583ef82f819c14c0/src/libs/core/CMakeLists.txt#L34-L36

It would probably be better to get brew's prefix with `$(brew --prefix)` or to predicate the prefix on the architecture.
I don't know enough CMAKE to do this myself unfortunately.

Cheers

zchrissirhcz:
A quick fix is
```bash
export LibArchive_ROOT=/opt/homebrew/opt/libarchive
cmake -S . -B build -G Ninja && cmake --build
```

trollixx:
Fixed in 1621.

PR #1607, opened 2024-03-16
build(cmake): correctly find libarchive on Apple-M series

zchrissirhcz:
For the issue https://github.com/zealdocs/zeal/issues/1602

### DIFF
--- a/src/libs/core/CMakeLists.txt
+++ b/src/libs/core/CMakeLists.txt
@@ -27,15 +27,21 @@ else()
     target_link_libraries(Core Qt5::WebEngine)
 endif()
 
-find_package(LibArchive REQUIRED)
-
-# Set LibArchive_INCLUDE_DIR on macOS when LibArchive is installed with Homebrew.
-# See https://github.com/Homebrew/legacy-homebrew/issues/21415.
-if(APPLE AND NOT LibArchive_INCLUDE_DIR)
-  set(LibArchive_INCLUDE_DIR "/usr/local/opt/libarchive/include")
+find_package(LibArchive QUIET)
+if(NOT LibArchive_FOUND)
+    find_path(LibArchive_INCLUDE_DIRS archive.h
+        PATHS /opt/homebrew/opt/libarchive/include /usr/local/opt/libarchive/include
+        REQUIRED
+    )
+    find_library(LibArchive_LIBRARIES
+        NAMES archive libarchive
+        PATHS /opt/homebrew/opt/libarchive/lib /usr/local/opt/libarchive/lib
+        REQUIRED
+        NO_DEFAULT_PATH
+    )
 endif()
 
-if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.17.0)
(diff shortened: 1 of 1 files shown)


zchrissirhcz:
I locally use a `mac-mini` with `brew install libarchive`, and this PR can fix build.

@alt-romes Can you please try it locally if possible?

zchrissirhcz:
@trollixx Could you please review this PR when you have a moment?

src/libs/core/CMakeLists.txt
```diff
@@ -27,14 +27,18 @@ else()
...
-find_package(LibArchive REQUIRED)
-
 # Set LibArchive_INCLUDE_DIR on macOS when LibArchive is installed with Homebrew.
 # See https://github.com/Homebrew/legacy-homebrew/issues/21415.
 if(APPLE AND NOT LibArchive_INCLUDE_DIR)
-  set(LibArchive_INCLUDE_DIR "/usr/local/opt/libarchive/include")
+  if(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+    set(LibArchive_INCLUDE_DIR "/usr/local/opt/libarchive/include")
+  elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
+    set(LibArchive_INCLUDE_DIR "/opt/homebrew/opt/libarchive/include")
```
trollixx, line 36: Is there a way to detect Homebrew installation rather than hard-coding the path? Also what if `arm64` has LibArchive installed not from Homebrew? I'd expect `find_package` to take care of this?

zchrissirhcz: Yes, using `find_package` looks more elegant. Let me do some trial for different LibArchive installation cases:
- Homebrew installed, both x86_64 and arm64
- Installed to system path, but not using homebrew
- User compiled and installed, but not installed to system path

zchrissirhcz: > Is there a way to detect Homebrew installation rather than hard-coding the path? Also what if `arm64` has LibArchive installed not from Homebrew? I'd expect `find_package()` to take care of this?

Actually there is a `libarchive.tdb` library file in subdirectories of macOS SDK. Previously I set `LibArchive_INCLUDE_DIR` to homebrew's one, then call `find_package()`, but the result is mismatched pair: homebrew installed headers, macOS SDK provided library file.

The new force-pushed commit now find the homebrew installed headers + library file.

zchrissirhcz: The installation of LibArchive, does not provide a `LibArchiveConfig.cmake` or `libarchive-config.cmake` file. It provides `libarchive.pc` file, which, can be used to find LibArchive headers and library files by invoking wrapper of `pkg-config`, but looks like not very simple.

trollixx:
@zchrissirhcz sorry I've been quite busy lately, was there an issue with the approach you proposed?

zchrissirhcz:
@trollixx The approach I proposed works. I sometime delete not-that-active forked repos. 
Now re-forked and re-PRed in https://github.com/zealdocs/zeal/pull/1621.

Issue #1655, opened 2024-09-11
Make docsets right-clickable to show a context menu


Afnankabiro:
A new **menu** will open when something in the docsets is **right clicked**. There would be some **options** in the menu.

I strongly suggest adding the option for **"Open in new tab"** to add here. Other options may be added in the future

Here is an example image-
![Untitled](https://github.com/user-attachments/assets/8b9410c0-625b-4aba-bc90-06347a9e2917)

PR #1673, opened 2024-12-06
Added Context Menu to Docsets

Roshni-P:
The newly added Context Menu, will allow the user to open a new tab from the Docset.

### DIFF
--- a/src/libs/ui/searchsidebar.cpp
+++ b/src/libs/ui/searchsidebar.cpp
@@ -16,6 +16,7 @@
 #include <registry/listmodel.h>
 #include <registry/searchmodel.h>
 #include <registry/searchquery.h>
+#include <ui/mainwindow.h>
 
 #include <QCoreApplication>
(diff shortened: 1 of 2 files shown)


src/libs/ui/searchsidebar.cpp
```diff
@@ -16,6 +16,7 @@
 #include <registry/listmodel.h>
 #include <registry/searchmodel.h>
 #include <registry/searchquery.h>
+#include <ui/mainwindow.h>
```
trollixx, line 19: This should be avoided. I'd rather have side bar emit signals that can be connected to in the main window.

trollixx (changes requested):
Having a menu with a single element is strange. Need to figure out more items to put there to justify adding it.

trollixx:
Thanks for contributing, but this has been sitting open for a while with no follow-up.

The original ask from #1655 still needs to be addressed.

PR #1689, opened 2025-04-02
Frontend changes to accommodate global docsets option

### LINKED ISSUE #516: Multiple docset storages
It might be convenient to have more than one docset storage.

Another use case would be have a global docset storage (e.g. in `/usr/local/share`) and the allow users to installed docsets personally (`~/.local/share`).

Some primitive conflict resolution should work fine.

lshprung:
Closes #516. Marking as draft because implementation is not done. The "use global docsets" option has been added to settings, but everything else is WIP.

### DIFF
--- a/src/libs/core/settings.cpp
+++ b/src/libs/core/settings.cpp
@@ -189,6 +189,7 @@ void Settings::load()
         docsetPath = QStringLiteral("docsets");
 #endif
     }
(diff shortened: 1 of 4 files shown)


trollixx:
Thanks for the PR. This feature is coming, but in a bit different way as I am trying to refactor the code to allow for more docset sources (and local storage paths) in general.

PR #1691, opened 2025-04-02
Installing dependencies

Sam4uk:
The script installs the libraries and tools needed for building

!!!TODO need to add dependencies for other systems. Fully implemented for Debian


trollixx:
Thanks for the contribution, but I don't think providing scripts for installing build dependencies provides much value. We have wiki that outlines build process for popular distributions. Maintaining a shell script would add unnecessary burden, especially given we cannot test it on all possible variations. Cheers.

PR #1758, opened 2026-02-21
Replace Q_UNUSED with [[maybe_unused]] attribute

stefanos82:
Since we are using C++17, specifically CMake uses gnu++17 to be exact, I thought it would be wise to modernize the code by using what the language offers us already, so we can reduce the code just a bit.

## Summary by CodeRabbit

* **Refactor**
  * Modernized internal code to use contemporary compiler annotations for unused parameters and updated type-alias usage across browser, UI, global-shortcut, registry, and core components.
  * No functional or behavioral changes; user experience and app behavior remain unchanged.


trollixx:
I prepared this exact change multiple times and never pushed. The code becomes so much less unreadable IMO.

stefanos82:
Feel free to close my PRs, no worries.

PR #1911, opened 2026-06-06
Add Copy-Selected operation to Zeal 

andrecbarros:
Add a simple Copy-Selection to Edit menu.

## Summary by CodeRabbit

* **New Features**
  * Added a "Copy Selection" action to the Edit menu with Ctrl+C keyboard shortcut for copying selected web content.


trollixx:
Thanks for the PR! Could you clarify what prompted this change? Does `Ctrl+C` not work for you when the web view has focus, or is the goal simply to make the copy action discoverable from the Edit menu?

andrecbarros:
The intention is just to make it discoverable and also accessible by mouse.

trollixx:
Thanks for the follow-up. Copy is already reachable here, both via `Ctrl+C` and the selection-aware Copy in the right-click context menu, so this is about menu discoverability rather than a missing capability.

The PR needs more work before it could land:

- It binds a custom, always-enabled `Ctrl+C` action that forwards unconditionally to the web view, ignoring which widget has focus. A selection in the search or Find-in-Page field would not copy correctly.
- The idiomatic approach is to reuse `pageAction(QWebEnginePage::Copy)`, which is already used for the context menu and auto-enables with the selection, rather than threading a new `copySelection()` method through three layers.
- A lone "Copy" is also an odd Edit-menu entry. If we add clipboard actions, the coherent set would be Copy plus Select All, also focus-aware.

The current implementation leaves too many gaps and even introduces new issues. I'll leave it open for now, but it would need a substantial rework to be considered. If you want to continue improving it, I think merging a comprehensive implementation would be a nice enhancement.

trollixx:
Per above.

PR #1912, opened 2026-06-08
feat(docsets): support compressed tarix docsets

### LINKED ISSUE #138: Compressed docset
Hi,

Do you have a plan to support compressed docset, like zipped archive? It will help very much in reducing the size and disk reading time.

### LINKED ISSUE #1278: C++ docset contains file with names that are invalid on Windows
(FYI - I'm using Zeal v0.6.1 on Windows 10)

The C++ docset contains a number of pages containing the character '*', which is invalid in Windows - for example, C++.docset/Contents/Resources/Documents/en.cppreference.com/w/cpp/utility/optional/operator*.html exists in the Dash C++ docset.

This means that navigating to pages like the one mentioned above doesn't work, because a) the URL is invalid on Windows, and b) the target file doesn't exist.

Is this something fixable? Seems to me that a general way to do it might be to modify (in some deterministic fashion) filenames containing illegal characters when extracting them from the downloaded docset, and using a `QWebEngineUrlRequestInterceptor` to dynamically redirect URL requests from the URL in the docset (which contains the illegal character(s)) to the modified file?

### LINKED ISSUE #1588: error 404 request path: <path>
Hello! Can you tell me, please, how to fix that?

trollixx:
Fixes #138.
Fixes #1278.
Fixes #1588.

## Summary by CodeRabbit

* **New Features**
  * Install and serve archive-backed (Tarix) docsets with deferred extraction and on-demand archived document reads.
  * UI: download/retry flow for compact-index (.tarix) files with “Install Anyway” fallback.

* **Improvements**
  * HTTP server: callback-driven mounts with MIME detection and safer mount/unmount lifecycle.
  * Docset metadata: detect/report archived docsets; registry mounts archived docsets via per-request handlers.

* **Tests**
  * Comprehensive tests for archive/index handling, long paths, and integrity verification.

* **Chores**
  * Added LibArchive/Zlib support for archive processing.

### DIFF
--- a/src/libs/core/extractor.cpp
+++ b/src/libs/core/extractor.cpp
@@ -3,6 +3,8 @@
 
 #include "extractor.h"
 
+#include <util/tarixarchive.h>
+
 #include <QDir>
 #include <QLoggingCategory>
 #include <QThread>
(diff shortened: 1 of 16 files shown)


tophf:
Are there any docsets that use tarix currently or is it planned? If not, maybe make Zeal auto-generate it from a downloaded tar.gz so that the unpacking happens only once and in-memory? Otherwise, how can I make a tarix docset myself?

trollixx:
> Are there any docsets that use tarix currently or is it planned? If not, maybe make Zeal auto-generate it from a downloaded tar.gz so that the unpacking happens only once and in-memory? Otherwise, how can I make a tarix docset myself?

I think all official and user-contributed Dash docsets include a tarix index, but cheat sheets don’t. The [docset generation guide](https://kapeli.com/docsets) doesn’t mention tarix, so I’ll work on updating [our wiki](https://github.com/zealdocs/zeal/wiki/Dash-Docset-Format) with that info. In the meantime, checking the code and the linked issue is probably your best option.

PR #1930, opened 2026-07-18
feat(ui): add a default zoom level setting

### LINKED ISSUE #1335: Persist the zoom level for the document browser
The default font size is a little bit small for me, so I need to zoom in. This works well, but the zoom level is lost on the next launch, so I have to zoom it again. It would be great if the level is saved and gets automatically applied upon next launch.

xevrion:
The document browser always started at 100%, so a preferred zoom had to be set again after every launch.

This adds a default zoom level to the Style group in Settings > Content. The value applies to newly created web views and is stored as a percentage, so it stays readable in the configuration file and survives changes to WebView::availableZoomLevels(). Zooming within a tab stays temporary, as before.

The zoom reset action and tooltip are renamed from "Actual Size" to "Reset Zoom", which is accurate now that the neutral level is configurable.

Refs #1335.

### DIFF
--- a/src/libs/browser/webview.cpp
+++ b/src/libs/browser/webview.cpp
@@ -7,6 +7,8 @@
 #include "webcontrol.h"
 #include "webpage.h"
 
+#include <core/application.h>
+#include <core/settings.h>
 #include <ui/browsertab.h>
 #include <ui/mainwindow.h>
 #include <ui/widgets/iconhelper.h>
@@ -23,13 +25,16 @@
 #include <QWebEngineSettings>
 #include <QWheelEvent>
 
+#include <limits>
+
 namespace Zeal::Browser {
 
 WebView::WebView(QWidget *parent)
     : QWebEngineView(parent)
 {
     setPage(new WebPage(this));
-    setZoomLevel(defaultZoomLevel());
+
+    applyDefaultZoom();
 
     // Enable plugins for PDF support.
     settings()->setAttribute(QWebEngineSettings::PluginsEnabled, true);
@@ -42,6 +47,31 @@ WebView::WebView(QWidget *parent)
 int WebView::zoomLevel() const
 {
     return m_zoomLevel;
+}
+
+void WebView::applyDefaultZoom()
+{
+    setZoomLevel(nearestZoomLevel(Core::Application::instance()->settings()->defaultZoomPercentage));
+}
+
+int WebView::nearestZoomLevel(int percentage)
+{
+    const QList<int> &levels = availableZoomLevels();
+    if (percentage < levels.first() || percentage > levels.last()) {
+        return defaultZoomLevel();
+    }
+
+    int nearest = defaultZoomLevel();
+    int minDelta = std::numeric_limits<int>::max();
+    for (qsizetype i = 0; i < levels.size(); ++i) {
+        const int delta = qAbs(levels.at(i) - percentage);
+        if (delta < minDelta) {
+            minDelta = delta;
+            nearest = static_cast<int>(i);
+        }
+    }
+
+    return nearest;
 }
 
 void WebView::setZoomLevel(int level)
@@ -87,7 +117,7 @@ void WebView::zoomOut()
 
 void WebView::resetZoom()
 {
-    setZoomLevel(defaultZoomLevel());
+    applyDefaultZoom();
 }
 
 QWebEngineView *WebView::createWindow(QWebEnginePage::WebWindowType type)
--- a/src/libs/browser/webview.h
+++ b/src/libs/browser/webview.h
@@ -25,6 +25,7 @@ public:
 
     static const QList<int> &availableZoomLevels();
     static int defaultZoomLevel();
+    static int nearestZoomLevel(int percentage);
 
     void zoomIn();
     void zoomOut();
@@ -38,12 +39,14 @@ protected:
     void contextMenuEvent(QContextMenuEvent *event) override;
 
 private:
+    void applyDefaultZoom();
     bool handleMouseReleaseEvent(QMouseEvent *event);
     bool handleWheelEvent(QWheelEvent *event);
 
     QMenu *m_contextMenu = nullptr;
     QUrl m_clickedLink;
-    int m_zoomLevel = 0;
+    // Invalid index, so the first setZoomLevel() is never skipped.
+    int m_zoomLevel = -1;
 };
 
 } // namespace Zeal::Browser
--- a/src/libs/core/settings.h
+++ b/src/libs/core/settings.h
@@ -56,6 +56,8 @@ public:
     int defaultFixedFontSize;
     int minimumFontSize;
 
+    int defaultZoomPercentage;
+
     enum class ExternalLinkPolicy : unsigned int {
         Ask = 0,
         Open,
--- a/src/libs/ui/settingsdialog.cpp
+++ b/src/libs/ui/settingsdialog.cpp
@@ -6,6 +6,7 @@
 #include "ui_settingsdialog.h"
 
 #include <browser/settings.h>
+#include <browser/webview.h>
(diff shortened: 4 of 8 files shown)


trollixx:
Thanks for the contribution! I think the feature is useful, but constantly saving zoom on every change would likely annoy many users, who only wanted to zoom in once and never needed it to persist. A better approach would be to add a "default zoom level" setting like most browsers already do.

xevrion:
thanks for the feedback, I'll push an update when I implement it.

src/libs/browser/webview.cpp
```diff
@@ -29,7 +31,11 @@ WebView::WebView(QWidget *parent)
     : QWebEngineView(parent)
 {
     setPage(new WebPage(this));
-    setZoomLevel(defaultZoomLevel());
+
+    applyDefaultZoomLevel();
+    connect(Core::Application::instance()->settings(), &Core::Settings::updated, this, [this]() {
+        applyDefaultZoomLevel();
+    });
```
trollixx: Rabbit is correct here. We should not override with current selections when default value changes. It should only apply to new tabs. Please drop the connection.

src/libs/browser/webview.cpp
```diff
@@ -87,7 +100,7 @@ void WebView::zoomOut()
 
 void WebView::resetZoom()
 {
-    setZoomLevel(defaultZoomLevel());
+    applyDefaultZoomLevel();
```
trollixx: The behavior itself is correct and matches what other browsers do, so please **do not** change this line.

The problem is that the two entry points still advertise the old meaning: the menu action is "&Actual Size" with the zoom-original icon in `mainwindow.cpp:464`, and the reset button tooltip in `browserzoomwidget.cpp:56` reads "Actual size". Neither is correct once the target is a configurable percentage.

Please rename both to "Reset Zoom" as part of this change.

src/libs/browser/webview.h
```diff
@@ -38,12 +38,13 @@ class WebView final : public QWebEngineView
...
 
 private:
+    void applyDefaultZoomLevel();
     bool handleMouseReleaseEvent(QMouseEvent *event);
     bool handleWheelEvent(QWheelEvent *event);
 
     QMenu *m_contextMenu = nullptr;
     QUrl m_clickedLink;
-    int m_zoomLevel = 0;
+    int m_zoomLevel = -1;
```
trollixx, line 49:
```suggestion
    // Invalid index, so the first setZoomLevel() is never skipped.
    int m_zoomLevel = -1;
```

src/libs/ui/settingsdialog.ui
```diff
@@ -510,6 +510,36 @@
...
+                <width>40</width>
+                <height>20</height>
+               </size>
+              </property>
+             </spacer>
+            </item>
+            <item>
+             <widget class="QLabel" name="label_18">
+              <property name="text">
+               <string>Default &amp;zoom level:</string>
```
trollixx: `&z` is already taken by `Si&ze` above. How about "Default zoom le&vel:"?

src/libs/ui/settingsdialog.ui
```diff
@@ -510,6 +510,36 @@
             </item>
            </layout>
           </item>
+          <item>
```
trollixx: Zoom is not really a property of font. I think it belongs in the Style group. Please put it at the bottom of that group.

src/libs/ui/settingsdialog.cpp
```diff
@@ -189,6 +194,8 @@ void SettingsDialog::loadSettings()
     ui->fontSizeComboBox->setCurrentText(QString::number(settings->defaultFontSize));
     ui->fixedFontSizeComboBox->setCurrentText(QString::number(settings->defaultFixedFontSize));
     ui->minFontSizeComboBox->setCurrentText(QString::number(settings->minimumFontSize));
+    const int zoomIndex = ui->defaultZoomComboBox->findData(settings->defaultZoomFactor);
+    ui->defaultZoomComboBox->setCurrentIndex(zoomIndex == -1 ? Browser::WebView::defaultZoomLevel() : zoomIndex);
```
trollixx:
```suggestion
    // Combo box indices mirror WebView::availableZoomLevels().
    ui->defaultZoomComboBox->setCurrentIndex(Browser::WebView::nearestZoomLevel(settings->defaultZoomPercentage));
```

src/libs/browser/webview.cpp
```diff
@@ -44,6 +50,13 @@ int WebView::zoomLevel() const
     return m_zoomLevel;
 }
 
+void WebView::applyDefaultZoomLevel()
+{
+    const int factor = Core::Application::instance()->settings()->defaultZoomFactor;
+    const qsizetype level = availableZoomLevels().indexOf(factor);
+    setZoomLevel(level == -1 ? defaultZoomLevel() : static_cast<int>(level));
```
trollixx:
```suggestion
    setZoomLevel(nearestZoomLevel(Core::Application::instance()->settings()->defaultZoomPercentage));
```

And introduce the following helper method:

```cpp
// webview.h
static int nearestZoomLevel(int percentage);

// webview.cpp
#include <limits>

int WebView::nearestZoomLevel(int percentage)
{
    const QList<int> &levels = availableZoomLevels();
    if (percentage < levels.first() || percentage > levels.last()) {
        return defaultZoomLevel();
    }

    int nearest = defaultZoomLevel();
    int minDelta = std::numeric_limits<int>::max();
    for (qsizetype i = 0; i < levels.size(); ++i) {
        const int delta = qAbs(levels.at(i) - percentage);
        if (delta < minDelta) {
            minDelta = delta;
            nearest = static_cast<int>(i);
        }
    }

    return nearest;
}
```

src/libs/core/settings.h
```diff
@@ -56,6 +56,8 @@ class Settings final : public QObject
     int defaultFixedFontSize;
     int minimumFontSize;
 
+    int defaultZoomFactor;
```
trollixx: Please rename this to `defaultZoomPercentage`, and the key to `default_zoom_percentage`. We already use "percentage" for this quantity in `WebControl::zoomLevelPercentage()`, while "factor" is what Qt calls the qreal ratio given to `setZoomFactor()`. Please also rename `WebView::applyDefaultZoomLevel()` to `applyDefaultZoom()`, so that "level" consistently indicates an index in `availableZoomLevels()`.

trollixx (changes requested):
Hi, thanks for reworking the PR. Please see my inline comments.

xevrion:
thanks, all are applied now, move the setting to the style group and switched the `nearestZoomLevel` and the percentage naming, renamed `Actual Size` to `Reset Zoom` and also dropped the settings connection so it only affects new tabs.

src/libs/ui/settingsdialog.ui
```diff
@@ -345,6 +345,36 @@
             </property>
            </widget>
           </item>
+          <item>
+           <layout class="QHBoxLayout" name="horizontalLayout_15">
+            <item>
+             <widget class="QLabel" name="label_18">
```
trollixx: A super minor nit. Just for better readability:
```suggestion
             <widget class="QLabel" name="defaultZoomLabel">
```

src/libs/browser/webview.cpp
trollixx: My comment here was poorly worded. Your change was correct and that's what I wanted to be kept intact, not the original line.

Please replace `setZoomLevel(defaultZoomLevel());` with `applyDefaultZoom();`.

trollixx (changes requested):
Great! Just two very minor nits and this will be good to go.

xevrion:
Both fixed, thanks.

trollixx (approved):

trollixx:
Thank you!

